PR #7, opened 2018-02-26
Update ssh.hbs

Rudloff:
I have to run this command every time on my Debian computer.


FlorentRevest:
Thanks for reporting. I assume this is not needed when using something like network manager or whatever.
This can be useful for other users anyway !

PR #12, opened 2018-05-20
Fix mailto format and email link for other contact

tambry:


FlorentRevest:
Good catch, thank you!

PR #13, opened 2018-05-27
Update ionicons to not horribly break site on iOS

daniellandau:
With the old ionicons the icons don't show on an iPhone, but instead take
~10000px of whiteness vertically making the site completely unusable.


FlorentRevest:
Great! Thanks a lot

PR #18, opened 2018-06-10
Add instructions to manually boot into Fastboot

M4GNV5:
The major benefit from this, is that people with a broken/unacessible Android Wear can still install AsteroidOS. It also means `adb` is only required for dual booting, not for installing.

I did some research and [created a list](https://m4gnus.de/watch/accessing_fastboot.txt) of how to access fastboot on all supported watches.
The most common variant requires the user to swipe from the top left of the screen to the bottom right. One could create a symbolic animation to explain this. There are some videos on youtube but i didn't want to embed/link to them e.g.: https://youtu.be/S77DSlgVVig?t=23s. However the explanation should be fine too.


M4GNV5:
Im closing this after the discussion in IRC.

The problem with this is temporary install requires adb so removing the tutorial on how to enable adb without having an interactive way for the user to select a temporary or permanent install beforehand doesn't make much sense.

Issue #19, opened 2018-06-22
Freedom & Privacy


kratz00:
... this is what it says on your homepage regarding the wearable devices.
Regarding privacy, why is it that your homepage does requests to google-analytics.com and ajax.googleapis.com. There some other requests to CDNs as well. All of them are probably not protecting the privacy of your visitors.

FlorentRevest:
I'm glad AsteroidOS has community members who care about their privacy as much as you do @kratz00 .
Thankfully the whole project is open to contributions, including this website. I would warmly welcome a pull request replacing the services provided by the CDNs we currently rely on.

PR #20, opened 2018-06-22
Removed google-analytics.com integration.

kratz00:
First pull request to improve user privacy (see issue #19).

### DIFF
--- a/templates/includes/javascripts.hbs
+++ b/templates/includes/javascripts.hbs
@@ -8,16 +8,3 @@
 <script src="//cdnjs.cloudflare.com/ajax/libs/holder/2.2.0/holder.min.js"></script>
 <script src="{{ assets }}/js/application.js"></script>
 
-{{#if site.analytics.google_site_id}}
-<!-- Analytics
-================================================== -->
-<script>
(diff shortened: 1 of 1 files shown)


FlorentRevest:
I'm afraid we may have had a misunderstanding @kratz00. With "replacing the services provided by the CDNs" I did not mean "suppressing the services provided by the CDNs" https://github.com/AsteroidOS/asteroidos.org/issues/19#issuecomment-399567464
We rely on analytics's dashboard to better understand what brings people to AsteroidOS and how we can help them better. If you think this service would be better served by a self-hosted system or a different company, feel free to propose an alternative with equivalent services but just suppressing a functionality is not satisfying.
As a simple reminder (I'm sure you already know about this), I'd also add that you can install addons to your browser to remove calls to google-analytics from the client side. If this is something you truly care about, you'd probably want to have this installed.

FlorentRevest:
Hi @kratz00. Sure that sounds like something we could do. I'll check this up with the server's admin and let you know if/when it happens.

kratz00:
Great :+1:

kratz00:
Cool, thanks for replacing google-analytics.com with your own/self-hosted solution. :+1:

PR #22, opened 2018-09-03
Removed leftovers of google-analytics.com integration.

kratz00:
Follow up of #20 
This commit removes the leftovers of the google-analytics.com integration, which has been removed in commit 7e470818df2e0e8e7441d8797d63f8304729a6fe


FlorentRevest:
This shouldn't be needed. As long as the site.analytics.google_site_id variable is not defined anymore, this block of code should not be integrated to the generated webpages...
Anyway, I guess if you really really want to make sure there is zero mention to Google, even in dead code, I can merge the PR, just in case...

PR #23, opened 2018-09-04
Privacy enhancements

kratz00:
These two commits should increase the privacy of your users.
This is a followup of issue #19 

There is still one external JavaScript on the _Contact_ page: https://kiwiirc.com/client/irc.freenode.net/#asteroid besides the Matomo one, I could find.


FlorentRevest:
Thank you @kratz00 ;) I took some time because I wanted to check whether the md5sum of your files would match the ones from upstream but it's alright.

PR #24, opened 2018-09-25
Typo fix

colinmahns:
It's an easy typo to make :)


FlorentRevest:
Oh, you taught me something then! :) Good catch, thank you!

PR #28, opened 2019-04-03
Correction

jrtberlin:
All the changed watches are the same device called "bluegill" by Compal. When opened up they have the required pinouts.

bluegill


FlorentRevest:
Great finding! :) maybe it would be good to specify that they have to be opened up to access the pinouts? And some information as to how to connect a USB cable to them (if you know how to)

jrtberlin:
I'm currently working on it

PR #30, opened 2019-05-14
Add instructions to build in Docker

T-vK:
I cleaned the instructions up a bit and added instructions on how to build AsteroidOS in Docker.

### DIFF
--- a/pages/wiki/building-asteroidos.hbs
+++ b/pages/wiki/building-asteroidos.hbs
@@ -3,42 +3,94 @@ title: Building AsteroidOS
 layout: documentation
 ---
 
-<p>If you decide to compile AsteroidOS from source be aware that it’s a simple process but the first build might take you a lot of time. Report any problem to the <a href="https://github.com/AsteroidOS/asteroid/issues">issues tracker</a>:</p>
+<p>If you decide to compile AsteroidOS from source be aware that it’s a simple process but requires a lot of disk space (potentially more than 100GB) and the first build might take you a lot of time (hours). Report any problem to the <a href="https://github.com/AsteroidOS/asteroid/issues">issues tracker</a>:</p>
 
 <video class="" id="" width="100%" height="100%" controls="" loop="" autoplay=""><source src="https://asteroidos.org/asteroid.mp4" type="video/mp4"></video>
 
 <div class="page-header">
   <h1 id="general-information">General information</h1>
 </div>
-These instructions have been tested on Ubuntu 19.04, but should also be applicable to Debian Sid (at least at the time of writing). For other distributions the dependencies may have different names and you may have to install additional ones.
+<p>You might want to build AsteroidOS in Docker because you'll get a clean build environment that works no matter what kinds of package repositories or package versions you have installed or how outdated your Linux distribution is or which Linux distribution you're using in the first place.</p>
+<p>Another advantage of Docker is that it <em>should</em> also work on Windows and OS X. (However, this has not been tested yet!)</p>
+<p>Using Docker also allows you to easily roll back. Consider the following: If you were to simply delete the repository and uninstall the `prerequisites` you will potentially remove packages that you had installed before and might actually still need.</p>
 
 <div class="page-header">
-  <h1 id="downloading">Downloading</h1>
+  <h1 id="git-setup">Clone the repository</h1>
 </div>
-<p>Start by installing the prerequisites:</p>
-<pre><code>apt-get install git build-essential cpio diffstat gawk chrpath texinfo python2 python3 wget shared-mime-info</code></pre>
 
-<p>If you haven't set a git name and email yet, please do so, otherwise the build will not succeed! For example:</p>
+<p>Clone the main repository using the following command:</p>
+<pre><code>git clone https://github.com/AsteroidOS/asteroid.git</code></pre>
+
+<p>If you haven't set a global git name and email yet, adapt the following <em>git config</em> commands to your information. (This is only required to clone some git repositories when building.)</p>
 
 <pre><code>git config --global user.email "you@example.com"
 git config --global user.name "Your Name"</code></pre>
 
-<p>Clone the main repository using the following command:</p>
-<pre><code>git clone https://github.com/AsteroidOS/asteroid.git
-cd asteroid/
-</code></pre>
-<p>This repository only contains a shell script that populates <em>src/</em> with OpenEmbedded and the appropriate Asteroid layers. Then, it setups the environment for a bitbake build. The following command will setup a build for <em>dory</em> (the LG G Watch) but you can also build an image for other watches by using the corresponding codename.</em></p>
-<pre><code>. ./prepare-build.sh dory # Be careful that this script must be sourced and not only ran
-</code></pre>
-<p>If the environment has been correctly setup, you should now be in the <i>build</i> subdirectory.</p>
+<p>Before you continue to <em>Build without Docker</em> or <em>Build with Docker</em>, make sure you're in the asteroid directory:</p>
+<pre><code>cd asteroid/</code></pre>
+
 <div class="page-header">
-  <h1 id="building">Building</h1>
+  <h1 id="build-without-docker">Build without Docker</h1>
+</div>
+
+<div class="page-header">
+  <h2 id="downloading">Downloading</h2>
+</div>
+<p>Install the prerequisites:</p>
+<pre><code>apt-get install git build-essential cpio diffstat gawk chrpath texinfo python2 python3 wget shared-mime-info</code></pre>
+
+<p>This repository basically only contains a shell script that populates <em>src/</em> with OpenEmbedded and the appropriate Asteroid layers. Then, it setups the environment for a bitbake build. The following command will setup a build for <em>dory</em> (the LG G Watch) but you can also build an image for other watches by using the corresponding codename. (Codenames can be found on the <a href="{{rel 'install'}}">Install page</a>.)</p>
+<pre><code>source ./prepare-build.sh dory # Be careful that this script must be sourced and not only ran
+</code></pre>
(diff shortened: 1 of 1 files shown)


pages/wiki/building-asteroidos.hbs
```diff
@@ -10,35 +10,91 @@ layout: documentation
 <div class="page-header">
   <h1 id="general-information">General information</h1>
 </div>
-These instructions have been tested on Ubuntu 19.04, but should also be applicable to Debian Sid (at least at the time of writing). For other distributions the dependencies may have different names and you may have to install additional ones.
+<p>These instructions have been tested on Ubuntu 19.04, but should also be applicable to Debian Sid (at least at the time of writing). For other distributions the dependencies may have different names and you may have to install additional ones.</p>
```
FlorentRevest: You should move this paragraph to inside the "build without docker" section

FlorentRevest: You should also add an introduction to why one could potentially prefer the Docker build over a build without Docker.

pages/wiki/building-asteroidos.hbs
```diff
@@ -10,35 +10,91 @@ layout: documentation
...
+<p>Remove the Docker container called <em>asteroidos-toolchain</em> in case it already exists:</p>
+<pre><code>sudo docker rm -f asteroidos-toolchain</code></pre>
+
+<div class="page-header">
+  <h2 id="docker-build-usage">Usage example</h2>
+</div>
+<p>Create a Docker container called <em>asteroidos-toolchain</em> from the Docker image which we also called <em>asteroidos-toolchain</em> and build AsteroidOS for <em>dory</em> (the LG G Watch) within the container:</p>
+<pre><code>sudo docker rm -f asteroidos-toolchain ; sudo docker run --name asteroidos-toolchain -it -v /etc/passwd:/etc/passwd -u `id -u`:`id -g` -v "$HOME/.gitconfig:/$HOME/.gitconfig" -v "$(pwd):/asteroid" asteroidos-toolchain bash -c "source ./prepare-build.sh dory && bitbake asteroid-image"</code></pre>
+
+<p>Note: The files created during the build simply go to the current direcotry (which should be <em>asteroid</em>), more spcifically, the output goes to the subdirectories <em>asteroid/src</em> and <em>asteroid/build</em>.</p>
```
FlorentRevest: directory*

FlorentRevest: specifically*

pages/wiki/building-asteroidos.hbs
```diff
@@ -10,35 +10,91 @@ layout: documentation
...
+<p>To build for a different watch than the LG G Watch, use its corresponding codename instead of <em>dory</em> when executing the <em>docker run ...</em> command.</p>
+<p>You can find the codenames for the supported watched on the <a href="{{rel 'install'}}">Install page</a>.</p>
+
+
+<div class="page-header">
+  <h3 id="docker-run-explained">What exactly does that long <em>docker rm ... ; docker run ...</em> command do?</h3>
```
FlorentRevest: I don't think we should have a separate section for that, you can just detail it in the above section
+ the wording of "What exactly does that long command do?" is not the wording I would expect to find in a documentation

pages/wiki/building-asteroidos.hbs
```diff
@@ -10,35 +10,91 @@ layout: documentation
...
+# -u `id -u`:`id -g`                      Ensures that the current user id and group id from the host is used on files inside the container to avoid permission issues.
+# -v /etc/passwd:/etc/passwd              Ensures that the user ids and groups from the host are also available in the Docker container. (Otherwise the -u and -g would be useless.)
+# /etc/passwd                             Contains the user names and their ids.
+# "$HOME/.gitconfig:/$HOME/.gitconfig"    Shares your user's git config with the container.
+# -v "$(pwd):/asteroid"                   Is used to mount the current directory (which is your asteroid git repo clone) into the container.
+
+# bash -c "source ./prepare-build.sh dory && bitbake asteroid-image"   This is the command to be executed inside of the container.
+                                                                       "dory" being the codename of your watch.
+                                                                       (You have to use source and not just "." here!)</code></pre>
```
FlorentRevest: I don't think it's necessary to tell users not to run a command that is not even given

pages/wiki/building-asteroidos.hbs
```diff
@@ -10,35 +10,91 @@ layout: documentation
...
 
 <div class="page-header">
   <h1 id="installing">Installing</h1>
 </div>
-<p>After a while, the generated image should be available in <em>build/tmp-glibc/deploy/images/dory/</em>.</p> Install AsteroidOS using your usual device's instructions.
+<p>After a while, the generated image should be available in <em>build/tmp-glibc/deploy/images/dory/</em>. (No matter if you built with or without Docker.)</p>
+<p>Install AsteroidOS using your usual device's instructions, which you can find when you go to the <a href="{{rel 'install'}}">Install page</a> and click your device there.</p>
```
FlorentRevest: "which you can find when you go to the Install page and click your device there"
->
"which you can find on the Install page."

pages/wiki/building-asteroidos.hbs
```diff
@@ -10,35 +10,91 @@ layout: documentation
...
+
+<div class="page-header">
+  <h2 id="docker-build-setup">Setup</h2>
+</div>
+
+<p>First install Docker and Git on your system if you haven't already, then clone the main repository using the following command:</p>
+<pre><code>git clone https://github.com/AsteroidOS/asteroid.git
+cd asteroid/</code></pre>
+
+<p>If you haven't set a global git name and email yet, please do so, otherwise the build will not succeed! For example:</p>
```
FlorentRevest: the "please do so, otherwise the build will not succeed!" is not a wording that should be found in a documentation. You should keep a neutral tone and it is implied that the instructions you give are necessary for the build to succeed, otherwise you wouldn't give them on the first place.
I would suggest "If you haven't set a global git name and email yet, adapt the following commands to your information." If you want you can also justify this with something like "This is only required to clone some git repositories when building" but no exclamation mark is necessary.

pages/wiki/building-asteroidos.hbs
```diff
@@ -10,35 +10,91 @@ layout: documentation
...
-<pre><code>git config --global user.email "you@example.com"
-git config --global user.name "Your Name"</code></pre>
-
 <p>Clone the main repository using the following command:</p>
 <pre><code>git clone https://github.com/AsteroidOS/asteroid.git
 cd asteroid/
 </code></pre>
-<p>This repository only contains a shell script that populates <em>src/</em> with OpenEmbedded and the appropriate Asteroid layers. Then, it setups the environment for a bitbake build. The following command will setup a build for <em>dory</em> (the LG G Watch) but you can also build an image for other watches by using the corresponding codename.</em></p>
+
+<p>If you haven't set a global git name and email yet, please do so, otherwise the build will not succeed! For example:</p>
```
FlorentRevest: See below for a comment on the same sentence.

pages/wiki/building-asteroidos.hbs
```diff
@@ -10,35 +10,91 @@ layout: documentation
...
+  <h1 id="build-with-docker">Build with Docker</h1>
+</div>
+
+<div class="page-header">
+  <h2 id="docker-build-setup">Setup</h2>
+</div>
+
+<p>First install Docker and Git on your system if you haven't already, then clone the main repository using the following command:</p>
+<pre><code>git clone https://github.com/AsteroidOS/asteroid.git
+cd asteroid/</code></pre>
```
FlorentRevest: Could you factorize the git clone part out of the with and without docker instructions? Since this is a shared first step

pages/wiki/building-asteroidos.hbs
```diff
@@ -10,35 +10,91 @@ layout: documentation
...
+
+<p>If you haven't set a global git name and email yet, please do so, otherwise the build will not succeed! For example:</p>
+
+<pre><code>git config --global user.email "you@example.com"
+git config --global user.name "Your Name"</code></pre>
```
FlorentRevest: You could actually also factorize those git commands out of the with/without docker sections, since it is also common.

pages/wiki/building-asteroidos.hbs
T-vK: Okay, sure I'll change it. What did you have in mind instead of `What exactly does that long <em>docker rm ... ; docker run ...</em> command do?`?  
(It sounds pretty okay to me to be honest.)

Would `Explanation of the above <em>docker rm ... ; docker run ...</em> command:` be better?

FlorentRevest: Yes it would be better

pages/wiki/building-asteroidos.hbs
T-vK: If you don't like it, I'll remove. But I thought it would be a good idea to mention it because in other places of the documentation `. ./prepare-build.sh` is used and it wasn't obvious to me that the `.` doesn't behave the same as `source` within a `bash -c "..."` call. For instance I could see please copying `. ./prepare-build.sh dory` and `bitbake meta-toolchain-qt5` from the `Creating an Asteroid App`-page.

FlorentRevest: Let's just replace . with source in the couple of remaining places. Half of the users miss the dot and complain that it doesn't work anyway.

T-vK:
Okay, I think I have adjusted everything you commented on. I also added `requires a lot of disk space (potentially more than 100GB)` which I think would be very good to know for everyone before attempting to build.

FlorentRevest:
The paragraph about Docker is a bit too dense to my taste but that will do.
Thank you ;)

T-vK:
No problem, you're welcome. :)

PR #31, opened 2019-05-14
Improve the README file

T-vK:
Since grunt will be installed to the node_modules, the right way to execute `grunt` would be `npx grunt`. 
Alternatively you could globally install `grunt`. However, that would be a a bit redundant.

So I just wanted to mention that the way to use grunt is `npx grunt`, unless you have an outdated npm version that doesn't come with npx yet. I also ended up some other minor things. (If you don't like them I can change it back.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,18 @@
 
 Assemble and Grunt are used to build the site. To get started:
 
-1. Clone this repository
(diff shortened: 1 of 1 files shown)


FlorentRevest:
Looks good to me, thank you

PR #33, opened 2019-05-14
Make code elements with lines longer than the width scrollable

T-vK:
Suggestion to have scrollbars in code tags with very long lines.

So ithat they would appear like this:

image

Instead of like this:

image

What do you think?


FlorentRevest:
Fair enough

T-vK:
Great :)

PR #51, opened 2020-02-10
Adds 'PineTime' to the impossible ports

jrtberlin:
The 'PineTime' gets a lot of porting requests and does not meet the hardware requirements. This PR clarifies the state in our public documentation.


nicolasmaia:
Unfortunately it doesn't clarify much. What is mean by unmet hardware dependencies? Which ones?

jrtberlin:
The microcontroller is under the minimum spec of Linux

PR #52, opened 2020-02-14
Correct compiler name to arm7vehf-neon

beroset:
Correct compiler name to match name specified in conf/distro/asteroid.conf


pages/wiki/creating-an-asteroid-app.hbs
```diff
@@ -3,7 +3,7 @@ title: Creating an Asteroid app
 layout: documentation
 ---
 
-<p>The app creation process of AsteroidOS needs a Software Development Kit generated by OpenEmbedded. You can either grab a prebuilt SDK for the nightlies <a href="https://release.asteroidos.org/nightlies/sdk/oecore-x86_64-armv7vet2hf-neon-toolchain-nodistro.0.sh">here</a> or for the latests stable release <a href="https://release.asteroidos.org/1.0/sdk/oecore-x86_64-armv7vet2hf-neon-toolchain-nodistro.0.sh">here</a> and install it on your system or you can build it yourself as follow.</p>
+<p>The app creation process of AsteroidOS needs a Software Development Kit generated by OpenEmbedded. You can either grab a prebuilt SDK for the nightlies <a href="https://release.asteroidos.org/nightlies/sdk/oecore-x86_64-armv7vehf-neon-toolchain-nodistro.0.sh">here</a> or for the latests stable release <a href="https://release.asteroidos.org/1.0/sdk/oecore-x86_64-armv7vehf-neon-toolchain-nodistro.0.sh">here</a> and install it on your system or you can build it yourself as follow.</p>
```
MagneFire: This link is actually broken now....

MagneFire:
I just looked at my local build. It seems that the original names are actually accurate... Is it possible that these names depend on the architecture of the watches that are currently supported?

If so, wouldn't it actually be better to add a note that these names may be different depending on the architecture of the watch?

pages/wiki/creating-an-asteroid-app.hbs
beroset: You're right.  There is something different about how this is being built on my machine.  I will investigate.

beroset: I have investigated and found that there was a problem on my installation.  Specifically, I had and old installation, did a git pull and rebuilt and inadvertently discovered a bug in git ( see https://bugzilla.redhat.com/show_bug.cgi?id=1803506 ).  Anyway, I deleted everything and rebuilt everything from scratch and the name now matches the original documentation.  I'm going to submit a request to revert the patch.  Thanks for questioning this, and my sincere apologies for the error.

MagneFire: No need to apologize things like this can always happen :smiley:. And good job on the investigation! Now we know that it was something related to your machine. Not some AsteroidOS specific thing.

PR #58, opened 2020-04-14
adds "Telescope" to FAQ

jrtberlin:
Adds "Telescope" + OpenStore link to the answer of "Where can I find a phone synchronization app?"


FlorentRevest:
Maybe we should mention Winsteroid too :) https://www.microsoft.com/en-us/p/winsteroid/9nd0r9rhngc2

jrtberlin:
@FlorentRevest last update: "2 years ago". Is it still maintained and works on the latest windows updates? cc: @thunderluca

PR #60, opened 2020-05-15
Remove white backgrounds of watch images.

MagneFire:
Mainly for people using extensions to darken bright pages.

image


FlorentRevest:
Cool :) what extension do you use? I never managed to get good results out of those extensions

MagneFire:
DarkReader, but it's not perfect.The Samsung Browser is the best, but that only works on Android :)

PR #61, opened 2020-05-29
Add some more info on sensor issues and tilt-to-wake known state

MagneFire:
This does not include how @hummlbach and I came up with a fix for sensors on sparrow. This should be done at a later stage.

The sensor information here was something I was stuck on for a long time while I was porting sturgeon.


FlorentRevest:
thanks :)

PR #74, opened 2020-08-31
make clear Huawei Watch GT/GT2 is not supported

LittleFox94:
Someone asked for that change in IRC, saw the question multiple times. Just make it clear on the website


FlorentRevest:
Could you use a message box and a full sentence? Something like https://github.com/AsteroidOS/asteroidos.org/blob/master/pages/install/sparrow.hbs#L29

FlorentRevest:
Thank you!

PR #79, opened 2020-10-07
converts porting status to markdown

jrtberlin:


MagneFire:
In general looks good :+1: 

Two changes I noticed over what it was:
- There are no horizontal lines between titles and content (Can be solved by adding: ***).
- The text for `Supported watches`, `Possible ports not supported yet`, `Impossible ports because of lack of pinouts` and `Impossible ports due to unmet hardware dependencies` is huge in comparison to the previous.

jrtberlin:
Thanks @MagneFire 
The horizontal lines are missing because my markdown preview plugin in vscodium stylizes headlines with a line below, so I didn't notice.
The smaller headlines have bad spacing. This can either be fixed with `&nbsp;` as in the PR or by changing the underlying theme.

MagneFire:
Hmm, looks like the lines are at the wrong location :wink: 
And it seems that the `&nbsp;` does work, but it also makes for weird left offset.

Before:
Screenshot from 2020-10-07 14-28-10
After:
Screenshot from 2020-10-07 14-27-56

I've tested some changes:
https://gist.github.com/MagneFire/fa8d5a5bb1ab9a7fe724279e955fd4b9

MagneFire:
Looks almost identical to what it was, with the exception of the `Supported watches`, etc titles.

I think this is good enough now :+1:

jrtberlin:
Thanks  @MagneFire

FlorentRevest:
Have you tried applying the page-header class to h1 titles ? I wonder if that would change other stuff, maybe we can make it specific to h1 titles in the documentation page

Issue #80, opened 2020-10-11
Change watch rating


MagneFire:
The recommended watches are the LG G Watch Urbane (bass) and the Huawei Watch (sturgeon).

Simply looking at the install page(https://asteroidos.org/install/) does not highlight this.

I think we should come up with a better method to distinguish the most recommended watches.

I purpose the following two changes:

#### Move to five star rating

Currently, there are watches that have the 3 stars while not being on-par with bass and sturgeon (harmony and smelt).
While others have 2 stars but have some essential features missing (swift, mooneye) others seems perfectly capable (wren, sparrow).

What do you think of the following:
1 star (horrible, has some serious issues): tetra
2 stars (ok, but missing some essential features (Bluetooth)): swift, mooneye
3 stars (good, but misses some great features (Always-on-Display, Tilt-to-wake, etc)): anthias, dory, (in)harmony
4 stars (Great, minor things are missing (microphone)): smelt, lenok, sparrow, wren
5 Excellent (No missing features): bass, sturgeon

#### Change watch order to star based

Instead of alphabetically ordering the watches. The best supported watches will be the first in the list, no need to scroll down.

Please let me know what you think of this. Other suggestions are also welcome :wink:

FlorentRevest:
I'm all in for that :)

jrtberlin:
I think both proposals are a good idea.

MagneFire:
As per: https://github.com/AsteroidOS/asteroidos.org/commit/1509f09e598ed83ba92d745437fb335a20a7e6f8 this has been realized.

PR #81, opened 2020-10-12
install: Change watch ratings and reorder.

MagneFire:
As discussed in https://github.com/AsteroidOS/asteroidos.org/issues/80

Also took the liberty to allow for markdown in content files :smile:

The watch stars is pretty much a copy of the current issue. Let me know if you agree with the current star rating. I based it on the current support matrix and the general support right now of the watches.

### DIFF
--- a/pages/main/install.hbs
+++ b/pages/main/install.hbs
@@ -3,73 +3,74 @@ title: Install
 slug: install
 ---
 
-<p>AsteroidOS can be installed as an alternative operating system on various smartwatches. Instructions for each model can be found below:</p>
-<p>If you have questions regarding the installation process, please check out the <b>FAQ</b> section.</p>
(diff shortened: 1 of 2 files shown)


FlorentRevest:
I'd give dory 4 stars. It used to be our best supported watch (before bass). The main issue is a race condition in the bluetooth initialization. Otherwise it works fairly well. The rest LGTM.

FlorentRevest (approved):

MagneFire:
The reason I gave `dory` 3 stars is because tilt-to-wake isn't working on it. But if it is otherwise good than increasing its stars to 4 is good too.

FlorentRevest:
Makes sense :) I'm fine with 3 stars for dory then

MagneFire:
ok, reverted my last commit back to what it was :smile:

PR #83, opened 2020-10-25
converts 'creating and asteroid app' into markdown

jrtberlin:


jrtberlin:
I decided to remove the link to the stable SDK since the stable version is outdated. The architecture changed and users would need to substitute the architecture in the instructions. I propose that we re-add the stable link once we have a new stable release.

PR #84, opened 2020-10-25
fixes watch image file extensions

jrtberlin:


MagneFire:
Sorry! Forgot to change these when I removed the background of the watch images.

PR looks fine :smile:

PR #85, opened 2020-11-03
install: Wrap support row when screen is too narrow.

MagneFire:
This PR changes the support table such that it wraps to the next row when it overflows.

Here is a comparison of what it looked like and what it looks like after the PR. Top one is the new implementation the bottom one is the old table.
asteroidosorg_support_wrap_prev_cur

Here are two images of what it looks like when the screen is too narrow:
asteroidosorg_support_wrap_demo

asteroidosorg_support_wrap_demo_2

Also I added a another color to indicate that the status of something is unknown:
asteroidosorg_support_wrap_color_palette

### DIFF
--- a/pages/install/anthias.hbs
+++ b/pages/install/anthias.hbs
@@ -4,23 +4,15 @@ deviceName: anthias
 layout: aw-install
 ---
 
-<table width="100%"><thead><tr>
-<th>Display</th>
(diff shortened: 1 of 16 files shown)


FlorentRevest:
Excellent :) thanks for also making easier to use the colors.

PR #86, opened 2020-11-03
bass: Add note about heart rate sensor.

MagneFire:
Please let me know if the note makes sense. Not sure if it's entirely clear what I am trying to say here.


FlorentRevest:
Eh, yeah it's unfortunate that we have to add this but I think it's good to have that documented. :)
Thank you MagneFire! :)

PR #90, opened 2020-11-25
enables markdown in the Gruntfile

jrtberlin:
I was checking #89 on a clean copy of the master branch to check for any breaking changes and noticed that it doesn't generate HTML files at all.

This commit adds the markdown file extension (`.md`) to the source file regex in the Gruntfile.


MagneFire:
Tested locally, your patch seems to solve this. :+1:

PR #95, opened 2021-01-25
Add instructions on how to get to fastboot

TheLastProject:


jrtberlin:
Hi, since `ADB` mode is not the default: Is there any reason you decided to document the `ADB` method instead of using `SSH`? 
`ssh root@192.168.2.15 'reboot -f bootloader'`

TheLastProject:
Well, ADB mode worked for me. I didn't try SSH mode. I can add the Developer Mode instructions too if you want.

jrtberlin:
> Well, ADB mode worked for me. I didn't try SSH mode. I can add the Developer Mode instructions too if you want.

Would be great if you could add this. Explaining both is probably good if one of them doesn't work for a user.
Thank you for improving the docs!

jrtberlin:
Thank you for improving our documentation!

PR #97, opened 2021-01-31
Update watch support state (sensors/aod)

MagneFire:
Updates the watch support state. Please confirm that I didn't make false claims :wink:


FlorentRevest:
Nice :) thank you for keeping this up-to-date. I'm not entirely sure about the mtk tilt to wake but it's best to underpromise than overpromise, we can always change the color later if this turns out to be incorrect.

PR #98, opened 2021-02-05
User used for app development has to be ceres

miathedev:
Thank you @MagneFire for your help. Pretty sure i would have spent many more hours if i did not had your help.

The user for running selfmade qt applications has to  be ceres. Root will not work.


MagneFire:
Good catch! Thanks for your contribution :smile:

PR #99, opened 2021-02-07
More app documentation

miathedev:
Add a complete documentation how to run a simple hello world application on your smartwatch.
I went through the complete documentation to test if theirs still something missing. Should be complete now.

Someone may pls go through my pull request and check for my English.

Even more is coming at a later state.
Like how to integrate a nice made app into the launcher or how to make a ipk out of your app to share it with friends.

### DIFF
--- a/pages/wiki/creating-an-asteroid-app.md
+++ b/pages/wiki/creating-an-asteroid-app.md
@@ -55,7 +55,7 @@ qtcreator
 
 This can be done automatically by prepending "source /usr/local/oecore-x86_64/environment-setup-armv7vet2hf-neon-oe-linux-gnueabi" before #!/bin/sh in _/usr/bin/qtcreator.sh_
 
-Now that you are in QtCreator go to ‘_Tools-&gt;Options-&gt;Devices_‘
+Now that you are in QtCreator go to ‘_Tools &#8594; Options &#8594; Devices_‘
 
 - Add a new Generic Linux Device.
 - Name it "AsteroidOS Watch".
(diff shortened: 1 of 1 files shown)


MagneFire:
Thank you so much for this PR. this is a huge improvement to our documentation!

I did notice one small thing. As you may know, we are currently upgrading the build system from `warrior` to `gatesgarth`. This has an important change on how apps are launched. We use `mapplauncherd` to improve performance of qml based apps, this is what the `invoker` command is used for. Apps are launched by opening them with the dlopen() function. This used to work for position-independent code. However a change in glibc made it so that only libraries can be opened that way.
The only solution I found is to compile AsteroidOS apps as libraries. This has the unfortunate side-effect that AsteroidOS apps can't be launched as an executable.

https://github.com/AsteroidOS/meta-asteroid/pull/57, see the mapplauncherd section.

So the question is: Can you change the `/home/ceres/apps/asteroid-helloworld` to `invoker --single-instance --type=qtcomponents-qt5 /home/ceres/apps/asteroid-helloworld` and verify that this works fine too? If this works you can also remove the `TODO` in the `Tips and tricks` section.

miathedev:
Im still allive, i just have some exams to finish. On 31.03 i have my last one.

I actually tried to compile your build @MagneFire and i failed doing so. After my last exam i will take a look why it failed on my machine.

MagneFire:
@miathedev Thanks for the update :smile:

All that stuff is now also merged. So, you should be able to use the regular https://github.com/AsteroidOS/asteroid/ source one too from now on if you run into further issues that is :wink:

dodoradio:
@miathedev What version of Qtcreator has the checkmark 'override deployment data from build system'? There does not seem to be a way to edit this in version 4.11, and must be edited in the [project].pro file

MagneFire:
Just a heads up, we are currently transitioning from qmake to CMake (https://github.com/AsteroidOS/qml-asteroid/pull/14). This might require further changes to the documentation.

miathedev:
Im going to recreate my fork. Stumbled upon some more things to add. And i have a branch conflicht!

PR #100, opened 2021-02-27
Add install instructions for the Huawei Watch 2.

MagneFire:
Update porting status page with mentions of the most relevant people who helped make this port possible.

I don't exactly know the stat of microphone, speaker and heart rate support. They are marked as unknown for now.

Please merge only after the Huawei Watch 2 images are available.


jrtberlin:
@MagneFire In case you still need an image for the Huawei Watch 2.

huaweiwatch2

MagneFire:
@jrtberlin Thanks, but I already have this one, slightly adjusted for the black screen instead of the WearOS screenshot:
https://github.com/MagneFire/asteroidos.org/blob/sawfish/assets/public/img/sawfish.png
But thanks for the reminder, I'll adjust the support table to what we know right now and merge this :smile:

MagneFire:
Ok, few changes since initial commit:
- Move slightly down in the install page, support isn't super good as the watch has the following quirks (some are fixed but not in upstream currently):
  - Speaker not working
  - Microphone not working
  - USB not working
  - UI not very smooth
  - Battery life is below average, even worse than WearOS
- Changed black screen watchface to be a little less black.

PR #103, opened 2021-04-14
Remove nowrap from inline code blocks

ChandlerSwift:
Before:
20210414_09h44m16s_grim

After:
20210414_09h49m04s_grim


FlorentRevest:
sweet! :)

PR #108, opened 2021-05-14
Add headline to section about getting toolchain

GeckoEidechse:
This change comes from me personally missing the link to the prebuilt SDKs and trying to build the toolchain from source simply due to starting to read the page from the first headline on.

Adding a headline above the first section explaining how to acquire the SDK and stating that building an SDK from source is an _alternative method_ should alleviate this problem.

I'm not 100% sure if this is correct as it's been a while since I fiddled with app development on AsteroidOS, so maybe someone with experience could OK this.


jrtberlin:
Thank you for the PR. 
I think the headline makes it more clear that a prebuilt SDK is an option. Next time you contribute it would be great if you could check for necessary empty lines beforehand.

PR #110, opened 2021-05-18
Add link to Discord

Doomsdayrs:
I created a discord server, Would be awesome to add it to the site footer.


jrtberlin:
Discord is a proprietary plattform with an unclear business model. We already have IRC/Matrix for communication. 
I don't think we should add an unofficial AsteroidOS channel to the official webpage.

jrtberlin:
It's great that you want to see the Asteroid community grow!

However, we came to the conclusion that It's not a good idea to split the small community up between multiple chats.
We also don't have a regular contributor on the project who is willing to actively use discord, and we don't want to list an unofficial channel on the webpage.

PR #114, opened 2021-06-12
added Timezone, date & time hint to useful-commands page

eLtMosen:
Thanks for review.

### DIFF
--- a/pages/wiki/useful-commands.md
+++ b/pages/wiki/useful-commands.md
@@ -85,11 +85,11 @@ systemctl --force reboot bootloader # Restarts the watch in bootloader mode
 systemctl --force reboot recovery # Restarts the watch in recovery mode
 ```
 
-# Timezone, date & time
+# Time zone, date & time
 
 ---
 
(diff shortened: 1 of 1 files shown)


pages/wiki/useful-commands.md
```diff
@@ -76,11 +76,23 @@ mcetool -M disabled # Disables the tap to unlock functionality
...
 reboot # Simply reboot your watch
 systemctl --force reboot bootloader # Restarts the watch in bootloader mode
 systemctl --force reboot recovery # Restarts the watch in recovery mode
 ```
+
+# Timezone, date & time
+
+---
+
+Although we offer [synchronization clients](https://asteroidos.org/wiki/synchronization-clients/) for different platforms, it might still be useful to set the time zone or synchronise the date and time using standard Linux tools. The Linux folder `/usr/share/zoneinfo/` contains the nameing scheme for your local time zone in <continent>/<zone> format.
```
jrtberlin, line 92: `nameing` → `naming`

`time zone in <continent>/<zone> format.` results in `time zone in / format.` since the `<words>` are interpreted as HTML tags. This needs either backticks or code tags to make it displayed correctly.

eLtMosen: Good spot :D
Sorry. 
Also what i see only now. What about `Timezone` vs `time zone` in title and text? Deepl says `Time zone` now when i check.

jrtberlin: I would be consistent and use `time zone`. In a headline you can use a consistent title case instead.

jrtberlin:
Thanks!

PR #118, opened 2021-08-10
Adapt SDK instructions for CMake.

MagneFire:
This PR finally addresses the lack of correct documentation for the new SDK that switches from `qmake` to `CMake`.

### DIFF
--- a/pages/wiki/creating-an-asteroid-app.md
+++ b/pages/wiki/creating-an-asteroid-app.md
@@ -52,8 +52,8 @@ This script will install the cross-compiler and ARM libraries in `/usr/local/oec
 
 ```
 source /usr/local/oecore-x86_64/environment-setup-armv7vehf-neon-oe-linux-gnueabi
-qmake
-make
+cmake -B build
+cmake --build build
 ```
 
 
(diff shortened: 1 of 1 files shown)


pages/wiki/creating-an-asteroid-app.md
```diff
@@ -75,34 +75,55 @@ Now that you are in QtCreator go to ‘_Tools-&gt;Options-&gt;Devices_‘
 - Add a new Generic Linux Device.
 - Name it "AsteroidOS Watch".
 - Choose 192.168.2.15 as IP address.
-- Use ceres as user.
+- Use root as user.
```
FlorentRevest, line 78: Ooc why does it need to be root now ? We should be able to run the app as user ceres. Is it maybe copying to an unwritable directory ? I thought the run command would push it to /tmp or something.

Otherwise, looks good to me! Thank you!

FlorentRevest (approved):

pages/wiki/creating-an-asteroid-app.md
MagneFire: Yes, the reason is indeed that QtCreator decides to push the binaries to `/usr/local/bin/` which requires root.
For the `invoker` to work I also needed to set the `XDG_RUNTIME_DIR` to `/run/user/1000` as otherwise it couldn't connect via D-Bus to the booster service...

FlorentRevest: Can we instead instruct it to put it in /tmp ? And run it as ceres?

MagneFire: That's what I initially did. But decided against it because it's very tedious to setup everytime you setup a new project.

@miathedev did also describe instructions for that in a previous PR https://github.com/AsteroidOS/asteroidos.org/pull/99/files#diff-e37cc1cf77fb59f01f81560400f6527c3f78f5f369765c373aad43c423f8f49cR115

That PR shows already part of the issue: you have to select the path to the locally built binary and a remote path. But even worse is that there appears to be no simple browse button for this purpose.

That's why I decided to leave this workaround as the solution. If you or @miathedev knows a better solution than that would be hugely beneficial as my solution is a bit of a hack in my opinion ;)

miathedev: Im sad that i dont have a better way yet.

Btw i cant await this PR to be merged. I didnt yet get CMake working. (And i didnt had any time yet to get more into it)

FlorentRevest: In the absence of a better alternative, it's fine to merge as is ;)

PR #123, opened 2021-10-30
Add Skipjack to supported devices list

R0NAM1:
As skipjack can currently run AsteroidOS (Although with very limited features at the moment) it is now a supported device, this PR is add it officially to the website and official build system.


jrtberlin (approved):
Thanks. I will merge your PR when a build is available to download!

MagneFire:
Overall, this looks good :+1: 

Only one comment I can find here is that `skipjack` should be positioned before `swift` as the list is alphabetically sorted based on number of stars and codename.

PR #129, opened 2021-11-14
Update emulator.md

### LINKED ISSUE #128: Add troubleshooting information for emulator
I had a hard time trying to get the emulator to compile and run correctly and discovered some things that may help others, such as how to resolve problems using QEMU with the proprietary NVIDIA driver.  This kind of information should be included in the emulator page.

beroset:
Add Troubleshooting section listing common problems and their solutions.  This fixes #128

### DIFF
--- a/pages/wiki/emulator.md
+++ b/pages/wiki/emulator.md
@@ -26,7 +26,7 @@ Unfortunately, most Linux distributions (such as Debian) still do not provide QE
 Once you have a correct version of QEMU, a rootfs and a kernel, you can start AsteroidOS in an emulator using:
 
 ```bash
-qemu-system-i386 -enable-kvm -kernel bzImage-qemux86.bin \
+qemu-system-x86-64 -enable-kvm -kernel bzImage-qemux86.bin \
 -device virtio-vga,virgl=on \
 -net nic -net user,hostfwd=tcp::2222-:22 \
 -drive format=raw,file=asteroid-image-qemux86.ext4 \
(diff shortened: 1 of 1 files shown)


pages/wiki/emulator.md
```diff
@@ -42,3 +42,41 @@ The previous command sets up a port forwarding so that you can connect to the em
...
+  -net nic -net user,hostfwd=tcp::2222-:22 \
+  -drive file=asteroid-image-qemux86.ext4,format=raw \
+  -m 512 \
+  -display gtk,gl=on \
+  -cpu qemu64,+ssse3,+sse4.1,+sse4.2 \
+  -usb -device usb-tablet -device usb-mouse \
+  --append "verbose root=/dev/sda rw mem=512M video=800x800"
+  ```
+  
+  ## Watch seems to lock up after a few seconds
```
FlorentRevest, line 77: I think you need to remove the space before ## for it to be considered a title ?

pages/wiki/emulator.md
```diff
@@ -42,3 +42,41 @@ The previous command sets up a port forwarding so that you can connect to the em
...
+
+## Video corruption using NVIDIA drivers
+There have been reported problems on machines using proprietary NVIDIA drivers.  One symptom is a black screen instead of the AsteroidOS screen which may or may not also show messages like these:
+
+```
+gl_version 46 - core profile enabled
+GLSL feature level 460
+failed to complete framebuffer 0x8cd7 asteroid-launch
+```
+Other symptoms include video artifacts (smearing, double screens) and crashing, usually with a message `Segmentation fault (core dumped)`.  If you're experiencing these and are using a proprietary NVIDIA driver, using `gtk` instead of `sdl` may work better.  An example command line:
```
FlorentRevest, line 63: What about making `-display gtk,gl=on` the default instruction ? :)

pages/wiki/emulator.md
```diff
@@ -42,3 +42,41 @@ The previous command sets up a port forwarding so that you can connect to the em
 ```
 ssh -p 2222 ceres@localhost
 ```
+
+# Troubleshooting
+Here are some common problems and their solutions.
+
+## `qemu-system-i386: warning: host doesn't support requested feature: CPUID.80000001H:EDX.lm [bit 29]`
+This just means that you are running an i386 emulation on a 64-bit host processor.  For AsteroidOS, this will make no difference, but you can run `qemu-system-x86-64` instead of `qemu-system-i386` to bypass the error, which has to do with [long mode](https://en.wikipedia.org/wiki/Long_mode) on x86-64 machines.
```
FlorentRevest, line 50: Nowadays everyone uses x86-64 machines anyway. I think it'd be better to make `qemu-system-x86-64` the default in the existing instructions

pages/wiki/emulator.md
FlorentRevest: Ooc, does it happen in normal conditions or is it a corner case ?

https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-nemomobile/mce/mce/0001-Keep-screen-on-by-default-on-emulator.patch should enable the demo mode by default on the qemu target

FlorentRevest:
Thanks a lot for investigating all of these!

Just a general comment regarding your commit messages (not just in this PR) When looking at the git log of a repository, commit titles such as "Update emulator.md" are not so informative. I'd recommend reading something like https://chris.beams.io/posts/git-commit/ this will be helpful for all sorts of open-source contributions, not just Asteroid :)

pages/wiki/emulator.md
beroset: Not a bad idea, but `sdl` works better if it works at all.  With `gtk`, the video is not quite as sharp, at least on the two machines I have.

FlorentRevest: Alright :)

beroset:
Improved the title for the log, replaced `qemu-system-i386` with `qemu-system-x86-64` everywhere and made small adjustments to style and content.

PR #130, opened 2021-11-21
Add more support options for skipjack and enable bluetooth

R0NAM1:
Since Skipjack now supports bluetooth, add a check to that, plus adding some other stuff like GPS and NFC,


R0NAM1:
Oh wow, that was fast!

PR #131, opened 2021-11-23
Add Sensor Support And Add Star To Skipjack

R0NAM1:
To be merged after https://github.com/AsteroidOS/meta-skipjack-hybris/pull/11 is merged, this PR adds Sensor Support where it counts, both in the Heart Rate Monitor and Tilt-To-Wake sensor, although Skipjack does not have a compass (Sad).

Also seems fitting based on the amount of supported hardware devices to bump Skipjack from two stars to three stars.


MagneFire:
Oopsie, sorry for not noticing this PR. Also good point to move the support to three out of five stars given that support is now on par to sawfish.

PR #134, opened 2021-12-12
Update WF creation guide with script explanations

eLtMosen:
As promised long ago, this is an extension of the watchface creation wiki to include all the findings we discussed in matrix chat and where found during creation of watchfaces over the last years.

Those where partly accumulated in the unofficial-watchfaces readme.md. Which this commit moves "here". https://github.com/eLtMosen/unofficial-watchfaces/commit/ee46a11f9cc030757ddaa14ec7a9768578bc3284

Detailed explanations of the script usage and some encouraging words to show how easy it actually is to get started, shall improve user and contributor flow into "self service" instead of coming to chat with redundant questions in light of possible 2.0 momentum.

Proof reading kindly supported by @beroset 
Any additional hints to include more details are very welcome.

### DIFF
--- a/pages/wiki/watchfaces-creation.md
+++ b/pages/wiki/watchfaces-creation.md
@@ -3,7 +3,7 @@ title: Watchfaces Creation
 layout: documentation
 ---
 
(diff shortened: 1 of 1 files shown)


FlorentRevest:
Very nice! Thank you you two :)

PR #135, opened 2021-12-13
Fix typos and fromat for better text flow

eLtMosen:
Its final011-masterdraft05-thoughtToBeReleaseWorthy05 now.
I think these are the last necessary corrections from my side.

### DIFF
--- a/pages/wiki/watchfaces-creation.md
+++ b/pages/wiki/watchfaces-creation.md
@@ -3,7 +3,7 @@ title: Watchfaces Creation
 layout: documentation
 ---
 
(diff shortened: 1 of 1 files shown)


pages/wiki/watchfaces-creation.md
```diff
@@ -3,7 +3,7 @@ title: Watchfaces Creation
...
+<p>Our goal is to make watchface creation fairly simple.<br>The provided scripts make it possible to create your first watchface even without any programming skills by redesigning existing SVG images.<br>Changing code requires some QML knowledge. Even if you are not familiar with QML yet, prior knowledge of any scripting language should be sufficiant to learn quickly from the already existing examples provided by the community.<br>Those community examples are collected in the <a href="https://github.com/AsteroidOS/unofficial-watchfaces">unofficial-watchfaces repo</a>. This guide will walk you through the steps of how to get started.</p>
```
FlorentRevest, line 6: Eheh, I find it funny how you went from "Watchface creation is fairly simple" to "Our goal is to make watchface creation fairly simple"

eLtMosen: Honesty wins.
And the "Is fairly simple" is/was obviously form your PoV :D
But what i see is that it gets easier and easier. So we can claim that as our goal easily. Which it is on my side.

PR #142, opened 2022-01-14
Move IP connectivity logic to a separate page

eLtMosen:
As hinted by user coolguest25 on matrix. 
Adding wifi instructions since most watches with wifi capability have it supported in the meanwhile.

### DIFF
--- a/pages/wiki/ip-connection.md
+++ b/pages/wiki/ip-connection.md
@@ -26,7 +26,7 @@ connmanctl&gt; connect wifi_CODE-FOR-YOUR-SSID
 connmanctl&gt; quit
 </code></pre>
 
-Verfify that you got an ip adress asigned with <code># ip a show dev wlan0</code>.
+Verfify that you got an ip address asigned with <code># ip a show dev wlan0</code>.
 
 Some more <a href="https://wiki.archlinux.org/index.php/ConnMan#Connecting_to_a_protected_access_point">details</a> on ArchWiki
 
--- a/pages/wiki/ip-connection.md
+++ b/pages/wiki/ip-connection.md
@@ -1,0 +1,49 @@
+---
+title: IP Connection
+layout: documentation
+---
+
+Configuring an IP connection on your watch has to be done manually until a GUI settings option is available.
+On watches with supported WLan, you can enable wifi and configure the connection using <code>connmanctl</code> like described below.
+Forwarding IP requests to a connected PC via USB is another option, explained in the second paragraph.
+
+By default, there is no <code>root</code> or <code>ceres</code> password, and no firewall rules. A password can be set using the <code>passwd</code> command.
+
+
+
+# IP over WLan
+
+*****
+
+Connect to your watch using <code>ssh root@192.168.2.15</code> or <code>adb shell</code>.
+
+<pre><code># connmanctl
+connmanctl&gt; enable wifi
+connmanctl&gt; scan wifi
+connmanctl&gt; services
+connmanctl&gt; agent on
+connmanctl&gt; connect wifi_CODE-FOR-YOUR-SSID
+connmanctl&gt; quit
+</code></pre>
+
+Verfify that you got an ip adress asigned with <code># ip a show dev wlan0</code>.
+
+Some more <a href="https://wiki.archlinux.org/index.php/ConnMan#Connecting_to_a_protected_access_point">details</a> on ArchWiki
+
+
+
+# IP over USB
+
+*****
+
+
+Configure a NAT **on your computer** (Note: Replace eth0 with the name of the interface that connects your computer to the Internet) with:
+
+    echo 1 > /proc/sys/net/ipv4/ip_forward
+    iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
+
+Configure a default gateway and DNS **on the watch** with the following commands ran via [SSH](https://asteroidos.org/wiki/ssh/):
+
+    route add default gw 192.168.2.1
+    echo "nameserver 8.8.8.8" >> /etc/resolv.conf
+
--- a/pages/wiki/ip-connection.md
+++ b/pages/wiki/ip-connection.md
@@ -4,14 +4,14 @@ layout: documentation
 ---
 
 Configuring an IP connection on your watch has to be done manually until a GUI settings option is available.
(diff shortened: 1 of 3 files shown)


FlorentRevest:
Let's extract the IP connectivity logic into a separate page and have both the SSH and Update pages refer to it.
This will avoid the kind of discrepancies we have right now between the pages. :)

pages/wiki/ssh.md
```diff
@@ -16,20 +16,6 @@ layout: documentation
...
-connmanctl&gt; services
-connmanctl&gt; agent on
-connmanctl&gt; connect wifi_<code for your ssid>
-connmanctl&gt; quit
-# ip a show dev wlan0
-</code></pre>
-<p>some more <a href="https://wiki.archlinux.org/index.php/ConnMan#Connecting_to_a_protected_access_point">details</a> on ArchWiki</p>
-<p>The watch should automatically get an IPv6 address from your network or you can set a static IPv4. You can disconnect from adb and connect via SSH.</p>
-<p>By default, there is no <code>root</code> or <code>ceres</code> password, and no firewall rules.</p>
+<p>Dropbear is already running on the watch so we just need to configured an IP address to connect.</p>
```
FlorentRevest, line 19: To configure* an IP address to connect to*

pages/wiki/update-asteroidos.md
```diff
@@ -3,70 +3,39 @@ title: Update AsteroidOS
...
-
-# OPKG over USB
+# Reflash AsteroidOS
 
 *****
 
+When no IP connection can be established on your watch, the easiest way to upgrade is to reflash the entire OS following the usual [installation instructions](https://asteroidos.org/install/).
 
-Configure a NAT **on your computer** (Note: Replace eth0 with the name of the interface that connects your computer to the Internet) with:
+You will first have to reboot into fastboot, which is slightly different under AsteroidOS than it is under Android Wear.
```
FlorentRevest, line 33: than it is under WearOS*

pages/wiki/ssh.md
```diff
@@ -16,20 +16,6 @@ layout: documentation
...
+<p>The <a href="https://asteroidos.org/wiki/ip-connection">IP Connection page</a> describes how to set up wifi.</p>
+<p>Once the setup is done, you can look for the IP adress your Router assigned via DHCP, with <code># ip a show dev wlan0</code>.</p>
```
FlorentRevest, line 21: Address*

pages/wiki/ip-connection.md
```diff
@@ -0,0 +1,49 @@
...
+
+Configuring an IP connection on your watch has to be done manually until a GUI settings option is available.
+On watches with supported WLan, you can enable wifi and configure the connection using <code>connmanctl</code> like described below.
+Forwarding IP requests to a connected PC via USB is another option, explained in the second paragraph.
+
+By default, there is no <code>root</code> or <code>ceres</code> password, and no firewall rules. A password can be set using the <code>passwd</code> command.
+
+
+
+# IP over WLan
```
FlorentRevest, line 14: This seems to always be written in all caps: WLAN https://en.m.wikipedia.org/wiki/Wireless_LAN (same comment for the other pages)
But I also alway had the feeling that this is a German word. In France for example no one understands what WLAN means, everybody talks about WiFi instead.

pages/wiki/ip-connection.md
```diff
@@ -0,0 +1,49 @@
...
+
+*****
+
+Connect to your watch using <code>ssh root@192.168.2.15</code> or <code>adb shell</code>.
+
+<pre><code># connmanctl
```
FlorentRevest, line 20: What’s the canonical way of integrating code blocks in these documentation pages ? Some seem to only have a `<code>`, some use `<pre><code>`, some use back ticks, some use indentation etc… maybe we should uniformize that a bit ?

pages/wiki/ip-connection.md
eLtMosen: WLAN is simply a LAN that uses radio technology instead of wiring to interconnect networked nodes. Which also includes Ad-hoc or completely proprietary implementations.
Wi-Fi is the wireless standard 802.11 and nothing else. 

WLAN not being generally understood is a shame imo and introduced the confusion we now have with 5G. Which is also just a WLAN by definition.
WLan is wrong and rooted in gamer community wanting to let the word look fancy like WiFi :D

I agree, we solely have Wi-Fi on our watches. I was coming from SEO perspective and tried to drop the WLAN term just for the search engines in case users search for "WLAN on Asteroid, etc" from outside.

eLtMosen: Also going for Wi-Fi instead if WiFi. 
WiFi makes sense from SEO perspective since it is top google choice. But all publications i trust use Wi-Fi officially.
Plus Wifi is already SEO covered by the `wifi` mentions we have.

pages/wiki/ip-connection.md
eLtMosen: I may be wrong but afaiu <pre> is for multi line <code> like ``` ?
The backticks where made possible by the move to markdown and i try to use them where possible.
But i have not fully understood where backticks/markdown works since e.g. here i hd to use <code> since backticks where not recognized *for some reason ®*

pages/wiki/ip-connection.md
```diff
@@ -0,0 +1,49 @@
...
+connmanctl&gt; enable wifi
+connmanctl&gt; scan wifi
+connmanctl&gt; services
+connmanctl&gt; agent on
+connmanctl&gt; connect wifi_CODE-FOR-YOUR-SSID
+connmanctl&gt; quit
+</code></pre>
+
+Verfify that you got an ip adress asigned with <code># ip a show dev wlan0</code>.
```
FlorentRevest, line 29: This adress slipped through

pages/wiki/update-asteroidos.md
```diff
@@ -3,40 +3,39 @@ title: Update AsteroidOS
...
 
 
-# Reflash AsteroidOS
+
+# Upgrade using OPKG
 
 *****
 
-The easiest way is to reflash the entire OS following the usual [installation instructions](https://asteroidos.org/install/).
+If you don't want to lose your personal data, and prefer upgrading Asteroid without reflashing it, you have two options to establish an internet connection from the watch.
```
FlorentRevest, line 14: Loose*

PR #155, opened 2022-03-29
ble-profiles: Reference the Apple Notification Center Service (ANCS) Specification.

MagneFire:
@FlorentRevest referenced this specification. It might be good to add it to our wiki too :smile:

### DIFF
--- a/pages/wiki/ble-profiles.md
+++ b/pages/wiki/ble-profiles.md
@@ -17,7 +17,7 @@ layout: documentation
   <h1 id="batteryprofile">Battery Profile</h1>
 </div>
 
-<p>Clients can read an Asteroid watch's battery level using the BLE battery profile. This profile is a standard <a href="https://www.bluetooth.com/specifications/gatt/viewer?attributeXmlFile=org.bluetooth.service.battery_service.xml">defined by the Bluetooth Consortium</a> very well documented and implemented by many other devices.</p>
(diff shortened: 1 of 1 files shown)


pages/wiki/ble-profiles.md
```diff
@@ -54,7 +54,7 @@ layout: documentation
   <h1 id="notificationprofile">Notification Profile</h1>
 </div>
 
-<p>Clients can create notifications on the watch side by using the Notification Profile. This profile is typically used to synchronize notifications from a phone to a watch or to send special alerts such as AsteroidOSSync's "Find my watch" feature.</p>
+<p>Clients can create notifications on the watch side by using the Notification Profile. This profile follows the <a href="https://developer.apple.com/library/archive/documentation/CoreBluetooth/Reference/AppleNotificationCenterServiceSpecification/Specification/Specification.html">Apple Notification Center Service (ANCS) Specification</a>. This profile is typically used to synchronize notifications from a phone to a watch or to send special alerts such as AsteroidOSSync's "Find my watch" feature.</p>
```
FlorentRevest, line 57: That's not correct.

We use our own custom Notification BLE profile on all platforms *except* for iOS where we use ANCS. It would be more appropriate to add a totally separate section where we explain that for iOS support specifically, we have the watch side of ANCS implemented.

PR #156, opened 2022-04-04
Add narwhal to install pages and porting status page

dodoradio:


pages/install/narwhal.md
```diff
@@ -0,0 +1,22 @@
...
+  <div class="support-col">Bluetooth<div class="support-col-good"></div></div>
+  <div class="support-col">Haptics<div class="support-col-good"></div></div>
+  <div class="support-col">USB<div class="support-col-good"></div></div>
+  <div class="support-col">WLAN<div class="support-col-good"></div></div>
+  <div class="support-col">Tilt-to-Wake<div class="support-col-good"></div></div>
+  <div class="support-col">Hands<div class="support-col-bad"></div></div>
+</div>
+
+<div class="callout callout-warning">
+    <p>Proper support for the hands will be added soon. A QML module is being worked on, which will allow applications to make use of the hands.</p>
```
jrtberlin, line 21: I would add that the hands are physical. Just to make it a bit more clear.

jrtberlin (changes requested):

pages/install/narwhal.md
dodoradio: I am still thinking about the change that would be most appropriate. I think that the module is actually close to being finished so mentioning it is not relevant. It would probably be better to write something along the lines of 

``this watch has a pair of hands embedded between the display and touchscreen panel. This obviously compromises the usability of the interface as the hands are a permanent obstruction. If you do not own one already, it is recommended to do a lot of research before buying.``

jrtberlin: That's alright imho.

dodoradio:
just squashed the commits to keep the history clean

jrtberlin (approved):

PR #158, opened 2022-04-12
More installation details & installation fix for carp/smelt on Windows

MagneFire:
Adds more details to the individual installation pages as a follow-up to https://github.com/AsteroidOS/asteroidos.org/pull/152.

Additionally, this uses a custom template for `carp/smelt` to refer to different drivers for Windows as the universal ones don't seem to work. This was reported by @docgalaxyblock (Thanks!)

### DIFF
--- a/pages/install/smelt.md
+++ b/pages/install/smelt.md
@@ -2,6 +2,7 @@
 title: Moto 360 2015
 deviceName: smelt
 repo: carp/smelt
+windowsDrivers: https://motorola-global-portal.custhelp.com/euf/assets/downloads/Motorola_Mobile_Drivers_64bit.msi
 section: install
 layout: aw-install
 ---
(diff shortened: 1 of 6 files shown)


pages/install/smelt.md
```diff
@@ -1,8 +1,9 @@
 ---
 title: Moto 360 2015
 deviceName: smelt
+repo: carp/smelt
 section: install
-layout: aw-install
+layout: aw-install-smelt
```
FlorentRevest, line 6: Instead of adding a new layout, add a new variable with a default value for the windows adb link and we can overwrite it for this watch

FlorentRevest (approved):

FlorentRevest:
Note for later: we should rename the template variable "repo" to better reflect that it's only used for as a hint for issues tags. But there's no reason to block this PR on this

PR #161, opened 2022-04-23
Add an env variable and troubleshooting section

beroset:
The CMAKE_PROGRAM_PATH variable was not set in the instructions and the result was that the generate_desktop CMake function would silently fail.  QtCreator also maddeningly refuses to allow the C and C++ compilers to be set to <No compiler> unless the Qt version (which is below those settings) is set to "None" first, so this was also documented. The troubleshooting section was added based on the recent experience of several new developers.


FlorentRevest:
Very nice, thanks for sharing your experience :)

PR #164, opened 2022-05-09
Change Wifi and Audio support matrix color to good/green

eLtMosen:
Tested on sparrow after https://github.com/AsteroidOS/meta-smartwatch/pull/64


MagneFire:
Awesome! :tada:

PR #173, opened 2022-07-27
add known install pages to porting status

docgalaxyblock:


beroset:
I certainly like the idea.  I wonder if it might make sense to also prominently feature the codenames, rather than only having them in the hidden URL?  It will also help with watches that are sold under multiple brands (Fossil Gen 4, for example is sold under Fossil, Diesel, and at least two others I don't recall right now).

eLtMosen:
Go from my side. Makes sense as an obvious help for users to provide a topic interlink here.

docgalaxyblock:
imo adding codenames or/and models to that side may be to much text - especially on mobile
maybe a table would be better then

FlorentRevest:
> I certainly like the idea. I wonder if it might make sense to also prominently feature the codenames, rather than only having them in the hidden URL? It will also help with watches that are sold under multiple brands (Fossil Gen 4, for example is sold under Fossil, Diesel, and at least two others I don't recall right now).

I agree with beroset here.

> imo adding codenames or/and models to that side may be to much text - especially on mobile
> maybe a table would be better then

Sure, a table could help

There is a conflict now because of MagneFire's recent PR sorry about that, since you'd have to change some things to fix the table would you mind also addressing the conflict ?

Also I find it a bit odd that `Samsung Gear Live *(unmaintained)*` is under `Supported watches and other ports` 

And `tetra` has a special note that it's not well supported but we now document this well on https://asteroidos.org/install/ and other watches which also aren't supported well (like minnow) don't have that note so I think it would be fair to drop the note about tetra in this page and just link to the install page that has the details.

Thanks for improving the doc, much appreciated!

docgalaxyblock:
I've fixed my previous accident 😅

FlorentRevest:
Ahah no problem, I see you haven't added a table or codenames yet so I'll wait until this is done before reviewing again.

docgalaxyblock:
iirc @beroset is work at improving/adding hardware overwiev to some page - therefore is the PR without a table
grafik
Should I add the codenames like that?
IMO at landscape phones that listing looks not so nice...

FlorentRevest:
Yes please add the codenames because they are important for our online communications. It's not important if documentation looks good or not on a phone, it's more important that documentation is useful than beautiful.

FlorentRevest:
Thank you!

docgalaxyblock:
@FlorentRevest Well you were a bit to fast
I add a minor italic+bold change to tetras *unmaintained* 🙈

PR #174, opened 2022-07-27
Dockerize

beroset:
This adds a Dockerfile and some description of how to use it to easily build and run a copy of the `asteroidos.org` documentation server within a locally hosted container.

PR #175, opened 2022-07-29
Add container docs

beroset:
This adds documentation to the web pages for the containerization implemented in https://github.com/AsteroidOS/asteroidos.org/pull/174 
They are technically independent, because they do not modify any of the same files, but it would make more sense to accept this one after #174 is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,38 +4,7 @@
 
 ## Quickstart
 
-Assemble and Grunt are used to build the site. To get started:
-
-1. Make sure you have the following installed:
-   * [NodeJS](https://nodejs.org/en/download/)
-   * [Git](https://git-scm.com/downloads)
-2. Run:
-   ``` Bash
-   git clone https://github.com/AsteroidOS/asteroidos.org.git # Clone this repository
-   cd asteroidos.org # Go to the root of the repo
-   npm install # Install dependencies
-   npx grunt # Build the site (If your npm version is so old that it doesn't support npx, run `npm i -g grunt` followed by `grunt` instead)
-   ```
-
-If all worked properly, you should have your website ready in an `_asteroidos.org` folder
-
-## Using `docker` or `podman`
-
-This can also be built and run inside a container even more simply.
-
-``` Bash
-podman build https://raw.githubusercontent.com/AsteroidOS/asteroidos.org/master/Dockerfile -t asteroidos-docserver
-```
-
-This will create an `asteroidos-docserver` image which is only around 236 MB and based on the Alpine Linux distribution.  To run it locally:
-
-``` Bash
-podman run -p8080:80 asteroidos-docserver
-```
-
-This will run a server in the container which can then be reached on your local machine by pointing your browser to `localhost:8080/`.  
-
-Although `podman` was used here, `docker` command syntax is identical.
+You can build and run a local copy of the web site on your own computer.  See https://asteroidos.org/wiki/localweb/ for details.
 
 ## License
 This website is a fork of lesscss.org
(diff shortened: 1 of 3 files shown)


pages/wiki/docserver.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Local containerized documentation server
```
FlorentRevest, line 2: More than a "documentation server" this is the asteroidos.org website

pages/wiki/docserver.md
```diff
@@ -0,0 +1,59 @@
...
+layout: documentation
+---
+
+This web site can be built and run within a software container, such as those used *container engines* such as [Docker](https://www.docker.com/) or [podman](https://podman.io/). One reason for doing this is to be able to install and build the documentation without having to load the collection of tools on your real computer. Building and running a documentation server in a local container is the topic of this page.
```
FlorentRevest, line 6: `such as those used *container engines* such as`
->
`such as those created by *container engines* like` ?

pages/wiki/docserver.md
FlorentRevest: Also I think we don't have to insist too much on this being containerized and just consider it as the default method to bring up the website.

So maybe "Local bringup of asteroidos.org" or something like this ?

pages/wiki/docserver.md
```diff
@@ -0,0 +1,59 @@
...
+podman build https://raw.githubusercontent.com/AsteroidOS/asteroidos.org/master/Dockerfile -t asteroidos-docserver
+```
+
+This will create an `asteroidos-docserver` image which is only around 236 MB and based on the Alpine Linux distribution. To run it locally:
+
+``` Bash
+podman run -p8080:80 asteroidos-docserver
+```
+
+This will run a server in the container which can then be reached on your local machine by pointing your browser to `localhost:8080/`.
```
FlorentRevest, line 23: So I have a tendency to think that if we document the same thing in two different places (both the asteroidos.org repo's README and the project doc) we will sooner or later have the two set of instructions diverge. wdyt ?

pages/wiki/docserver.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Running with modifications
```
FlorentRevest, line 25: This looks like the same thing repeated a third time now, maybe you can consolidate both sections ?

It's very likely that anyone bringing up asteroidos.org locally does it to introduce modifications so it's fair to consider this the default workflow

pages/wiki/docserver.md
beroset: I agree.  Maybe it should just be on the web site and the README can point to it?

pages/wiki/docserver.md
beroset: That's a good point.  I'll combine then and reduce the duplication.

pages/wiki/docserver.md
beroset: Better.  I'll make that change.

pages/wiki/docserver.md
beroset: Maybe "local containerized asteroidos.org"?  I thought it was important to note that it's containerized so it doesn't, say, require a user to install and configure Apache. Maybe that's too much to try to convey in just the title?

FlorentRevest: I personally don't think we should stress "containerized" in the title because most of the users don't really care whether it's containerized or not, they just want to bring up the website for local development.

I think we should just document the container way as the only way to bring up the website because it bboth simplifies work on the website and also keeps npm's pile of trash under some amount of control.

pages/wiki/docserver.md
FlorentRevest: Sgtm

pages/wiki/docserver.md
beroset: OK, I'm going to go with "Local version of asteroidos.org" if that's OK with you.

pages/wiki/docserver.md
beroset: Should I then also remove the original build instructions from the README.md and just point to the web site?  Seems like that might be cleaner and clearer.

pages/wiki/docserver.md
FlorentRevest: yep

pages/wiki/docserver.md
FlorentRevest: yep

FlorentRevest:
Neat! :)

Issue #176, opened 2022-08-09
Consolidate watch feature information


beroset:
The feature list for each watch is listed under the `Install` page for that watch.  This is unwieldy and repetitive and does not easily allow for such things as a comparison page that shows all watches and features.  

As a first step, it would be useful to gather all of the feature data into a `.json` file and then use that file with a helper to populate the existing pages.  

The next step will be to create a page showing all watches and all features, perhaps allowing for the ability to select features of interest and filter watches based on that.

PR #177, opened 2022-08-09
Restructure watch features

beroset:
Address part one of #176 by creating and using a json file that contains all watch models and features.  It also removes some unintended inconsistencies in the various install template files.


FlorentRevest:
Yaay :) Super cool, thanks a lot, that's a great improvement!

Just a general thing first: "features.json" is a bit of a generic name, when navigating pages it wouldn't immediately be obvious what that means (it may sound like a website feature). Maybe watches-support.json would be better suited ?

And then a nit, don't feel that you have to fix it because this is already a great improvement and I'd merge it as is (with the new name) but I see you define the supported and experimental array in the main-install page, but while we are at it why not express this distinction in the json model and have a helper function look it up, this way it would really be effortless to add a new watch to the website :) 

Anyway, thanks again for improving this kind of stuff

beroset:
There are more improvements that could be made, but this is now feature complete, tested, and includes the features table with all watches and all features.

beroset:
Rebased with current master.

beroset:
Revised commit based on matrix conversation -- revert to "Install" rather than "Watches".  Later we can add a "Watches" top layer to describe watches features.

pages/watches-support.json
```diff
@@ -0,0 +1,434 @@
...
+        { "name":"Touch", "status":"good" },
+        { "name":"Microphone", "status":"bad" },
+        { "name":"Bluetooth", "status":"bad" },
+        { "name":"Compass", "status":"good" },
+        { "name":"Haptics", "status":"good" },
+        { "name":"USB", "status":"good" }
+      ]
+    },
+    {
+      "name":"narwhal",
```
docgalaxyblock, line 220: `narwhal` has a working compass

pages/watches-support.json
```diff
@@ -0,0 +1,434 @@
...
+        { "name":"Compass", "status":"good" },
+        { "name":"Haptics", "status":"good" },
+        { "name":"USB", "status":"good" },
+        { "name":"WLAN", "status":"good" },
+        { "name":"Tilt-to-Wake", "status":"good" },
+        { "name":"Always-on-Display", "status":"good" }
+      ]
+    },
+    {
+      "name":"wren",
```
docgalaxyblock, line 416: tilt-to-wake works on `wren`

pages/watches-support.json
```diff
@@ -0,0 +1,434 @@
...
+        { "name":"Bluetooth", "status":"good" },
+        { "name":"Haptics", "status":"good" },
+        { "name":"USB", "status":"good" },
+        { "name":"WLAN", "status":"good" },
+        { "name":"Heart Rate", "status":"good" },
+        { "name":"Tilt-to-Wake", "status":"good" }
+      ]
+    },
+    {
+      "name":"sparrow",
```
docgalaxyblock, line 322: `sparrow` has a working speaker

MagneFire:
@docgalaxyblock Thanks for your feedback on this PR!
However, I think it might be better to leave these feature changes to a followup PR, once this is merged as this PR already contains quite a bit of  changes.
It's also the reason I marked my PR (https://github.com/AsteroidOS/asteroidos.org/pull/179) as draft, I'll update that one once this one is merged.

Of course let me know if you disagree with this! I just think it's important to prioritize the structure of the data rather than correcting the data with this PR too.

FlorentRevest:
Awesome work, thanks a lot beroset!

And I agree with what MagneFire brought up, let's address these separately :)

PR #178, opened 2022-08-13
Security fixes

beroset:
This updates a number of npm modules and replaces some obsolete ones.  It's technically independent, but it looks better (closer to the existing web site) if it's applied after #177.  This reduces the number of security flaws identified by `npm audit` from 116 to 42, although because the relevant npm modules are only executed during the assembly phase and not run when users access the site, there is not really any security concern.


FlorentRevest:
Woohoo thanks a lot!

Should package-lock.json be in gitignore? (Can be a separate PR naturally)

PR #184, opened 2022-08-30
Remove obsolete sturgeon heartrate monitor warning

beroset:
Since issue #82 is now fixed and closed, the warning about the heartrate
monitor not working on sturgeon is now obsolete.  This patch removes
that warning.


beroset:
To be clear, the issue #82 referred to above is this one:  https://github.com/AsteroidOS/meta-smartwatch/issues/82

PR #185, opened 2022-08-30
Maintainers to json

beroset:
This is the final consolidation of watch info for #176 which adds the list of maintainers to the json file and restructures the porting status page to use data from the json file rather than duplicating it.

### DIFF
--- a/pages/watches-support.json
+++ b/pages/watches-support.json
@@ -6,6 +6,7 @@
         "Asus Zenwatch 1"
       ],
       "status":"supported",
+      "maintainers": [ "locusf" ],
       "stars":3,
       "features": [
         { "name":"Display", "status":"good" },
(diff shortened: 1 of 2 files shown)


pages/watches-support.json
```diff
@@ -400,6 +419,7 @@
         "Sony Smartwatch 3"
       ],
       "status":"experimental",
+      "maintainers": [ ],
```
MagneFire, line 422: @FlorentRevest was the original maintainer.
I think marking this watch as *unmaintained* again is not right since it had recent work that improved support (battery life fix).

MagneFire (commented):
Thanks for this PR!
Looks great in general.

Only one minor comment, but I guess it's debatable if we want to mark `tetra` as *unmaintained* still. I know that it's probably just copied from the current state. So, I'm fine with merging this as is too.

MagneFire (approved):

FlorentRevest:
I love these refactorings :) keep them coming!

PR #186, opened 2022-09-06
Tell users to set time before using opkg

beroset:
It's common for users to report a problem trying to update with opkg
that is a result of not having the correct date and time set on the
watch before using opkg so that the watch can validate the SSL peer
certificate of the server.  This adds that note to the instructions and
also shows the error message that results if one fails to follow them.

### DIFF
--- a/pages/wiki/update-asteroidos.md
+++ b/pages/wiki/update-asteroidos.md
@@ -9,11 +9,11 @@ As currently no AsteroidOS watch has access to the internet by stock means, no G
 
 # Upgrade using OPKG
 
-*****
+Set the correct time and date on the watch. AsteroidOS needs to validate the remote server's certificate and establish a secure connection.
 
-If you don't want to loose your personal data, and prefer upgrading Asteroid without reflashing it, you have two options to establish an internet connection from the watch.
+If you don't want to lose your personal data, and prefer upgrading Asteroid without reflashing it, you have two options to establish an internet connection from the watch.
 
-In case your watch supports WLAN, connect it to your local Wi-Fi network using <code>connmanctl</code> as described on the [IP Connection page]({{rel 'wiki/ip-connection/). Alternatively you can share your internet connection from a PC to the watch via USB. '}}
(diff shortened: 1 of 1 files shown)


pages/wiki/update-asteroidos.md
```diff
@@ -9,11 +9,11 @@ As currently no AsteroidOS watch has access to the internet by stock means, no G
 
 # Upgrade using OPKG
 
-*****
+The first step in using `opkg` is to make sure the time and date are set correctly.  That's required so that the watch can correctly validate the remote server's certificate and establish a secure connection.
```
eLtMosen, line 12: Maybe a bit shorter and put the instruction first. Correctly is used two times. The "so" and "That's" are colloquialisms we can avoid in docs imo.

"Make sure the time and date are set correctly before using `opkg`. Else the watch can not validate the remote server's certificate and establish a secure connection".

eLtMosen (commented):
Good idea to add the info in two places!
If a user skips through and misses the first advise to set time. There is a common problems section catching the problem.
And again, very nice of you realising and solving this issue. Indeed many users ask that question.

pages/wiki/update-asteroidos.md
eLtMosen: Great, even shorter! But now it could mean to set the time and date on the host (with a lot of stretch and noob factor)?
How about
"Set the correct time and date on the watch. AsteroidOS needs to validate the remote server's certificate to establish a secure connection."

beroset: Perfect.  That's what I'll use.

eLtMosen: It took us 20 minutes to polish that sentence. But with thousands of users now reading one second less. We made the world a little tad better :rofl: 
Not to think of the massive time saver compared to when all those would have run in the issue and ask for help.
Again, nice work to pick this up.

eLtMosen:
Thanks for correcting my loose typo btw.
I figure, as a regular you count as trusted party by now. Two party principle met. Merging quickly :smile_cat:

PR #190, opened 2022-09-22
Specify only C2 and C2+ have NFC

Commenter25:
Extremely nitpicky change. The S2 doesn't have NFC, so this just makes that distinction.

### DIFF
--- a/pages/watches-support.json
+++ b/pages/watches-support.json
@@ -315,7 +315,7 @@
         { "name":"Microphone", "status":"bad" },
         { "name":"Bluetooth", "status":"good" },
         { "name":"GPS", "status":"bad" },
-        { "name":"NFC", "status":"bad" },
+        { "name":"NFC (on C2/C2+)", "status":"bad" },
         { "name":"WLAN", "status":"bad" },
         { "name":"Heart Rate", "status":"good" },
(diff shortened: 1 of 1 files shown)


docgalaxyblock:
@Commenter25 I am currently resorting the list in #188 
I could add that but I guess I would use a note outside of the table

Commenter25:
It seems like that would be hidden in the source code? My intent was to have something user-facing. If someone doesn't own a compatible watch, but wants to try Asteroid, the compatibility list is essentially their shopping list. So I think for the sake of clarity, model discrepancies like that should be displayed. S2 and C2(+) (and i'm willing to bet the E2) run off the same hardware, and work off the same image, but have different hardware features. I imagine a scenario in my head where someone sees a hardware group with the features they want, goes to ebay and buys a watch, then realizes the specific model they got doesn't have something on a hardware level. Though, perhaps there is a better way to implement this way of viewing it. There seems to be a lot of discussion on Matrix about how to organize watches, so I guess consider this me throwing my hat in the ring :p

docgalaxyblock:
Right, currently I am using the json as a json and my note list.
I planned to use the big yellow `Note` feature to mention hardware differences since the table gets very obstructive on mobile
WIP soon:tm:

PR #191, opened 2022-09-27
Unify similar watches.

MagneFire:
This PR unifies similar watches into a single page.
For now I left the install page the same by using the `reference` keyword in the watches model.

The main goal of this PR is to shed some light into the different Fossil Gen 4 watches. Which is achieved by unifying `ray` and `firefish` but providing a warning text box with information regarding the different versions (similar to `sparrow` and `wren`).

Preview:

https://user-images.githubusercontent.com/7857908/192627328-c8365138-60f3-4ccb-b97c-8a042e26a36c.mp4

This should also fix https://github.com/AsteroidOS/meta-smartwatch/issues/48.


eLtMosen:
This looks great and nicely consolidates information. Also it visually confirms implicit that those models are similar. Which might not be apparent to new users. Good Job!

pages/install/firefish.md
```diff
@@ -1,16 +1,37 @@
...
+       <li>Misfit Vapor 2 (46mm)</li>
+    </ul>
+    <code>ray</code> is compatible with the following variants:
+    <ul>
+       <li>Fossil Gen 4 (41mm)</li>
+       <li>Skagen Falster 2</li>
+       <li>Fossil Q Venture HR</li>
+       <li>Misfit vapor 2 (41mm)</li>
+    </ul>
+</div>
```
dodoradio, line 29: there's a few more models, such as the `michael kors access runway` (which seems to be a ray) and some models from guess connect that I can't confirm the names of (the current source for this is the wearos spec spreadsheet, but it's not reliable)
Imo it's worth saying that other models may also be compatible. There's a lot of them and information on some of them is rather flaky.
It might also be worth mentioning that firefish generally has model number DW6xx and ray has DW7xx, as a way of identifying whether the user's watch is a ray or a firefish (most of the watches' names are only printed on the box and the caseback only has the brand name so are hard to identify)

pages/install/firefish.md
```diff
@@ -1,16 +1,37 @@
...
 
 <div class="callout callout-info">
     <h4>USB connection</h4>
     <p>The Fossil Gen 4 doesn't expose a USB connection. It is however possible to connect to the watch using an adapter cable.</p>
     <p>You will need access to the USB pads on the watch, this can be achieved by removing the plastic back cover. Using heat and a prying tool, it is fairly straightforward to remove. Pay attention to the flex cable that connects to the back cover to the topside of the PCB. It is recommended to start prying on the side opposite of the buttons.</p>
-    <p>Then you can solder wires to connect a USB cable to the watch. The image below indicates the function of each pad.</p>
+    <p>Then you can solder wires or use a custom dock to connect a USB cable to the watch. The image below indicates the function of each pad.</p>
```
dodoradio, line 34: I'd put in a warning against soldering, as the pads are small and fragile - it's easy to mess this up and it's worth building some kind of dock

pages/install/sparrow.md
```diff
@@ -1,12 +1,13 @@
...
-layout: aw-install
+deviceNames: [sparrow, wren]
+label: sparrow/wren
+layout: aw-install-multi
 ---
 
 <div class="callout callout-warning">
     <h4>Warning!</h4>
-    <p>Make sure you use the correct Asus Zenwatch 2 ROM.</p>
+    <p>Make sure you use the correct Asus Zenwatch 2 image.</p>
```
beroset, line 10: I kept meaning to fix this but didn't ever remember.  Nice catch!

beroset (approved):
Definitely an improvement

FlorentRevest (commented):
LGTM overall, good to merge when @dodoradio's comments will be addressed :)

pages/install/firefish.md
MagneFire: Added a warning text below: "Be careful when soldering wires to the pads as the pads on the watch are very fragile."

pages/install/firefish.md
MagneFire: I've added the Michael Kors Access Runway and removed the Fossil Gen 4 lines as they are redundant with the other Fossil models.

MagneFire: Also added the DW6xx and DW7xx.

MagneFire:
Updated the PR with your suggestions, this is what it looks like right now:
Screenshot from 2022-10-02 22-15-12

PR #192, opened 2022-09-29
Add support for "partial" as a status

beroset:
The currently supported status categories for watch features are "good", "bad", "unknown", and "na", but a discussion on matrix had multiple people asking about the possibility of a "partial" status.  This changes the yellow-orange color previously used for "unknown" to signify "partial" and uses a blue color for "unknown".


MagneFire:
Looks good to me and it makes it at least clear that support is not in between unsupported and supported.

PR #196, opened 2022-10-16
featurelist: Fix issue where unified watches would link to old install pages.

MagneFire:
https://github.com/AsteroidOS/asteroidos.org/pull/191 has merged similar watches into a single page. This meant that the old pages are no longer valid and from now on they should refer to the merged page.

This fixes the issue where https://asteroidos.org/install/features/ would still link to `wren` and `ray` pages even though they are no longer valid.

This bug was reported by @docgalaxyblock


beroset (approved):
Good fix.  I've tested this.

PR #201, opened 2022-10-30
[building-asteroidos.md] Add Arch package install

eLtMosen:
Add prerequisites for Arch Linux

### DIFF
--- a/pages/wiki/building-asteroidos.md
+++ b/pages/wiki/building-asteroidos.md
@@ -48,9 +48,9 @@ Install the prerequisites:
 
 | Distro | Command |
 | ------ | ------- |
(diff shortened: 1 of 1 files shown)


beroset (changes requested):
If the packages were listed in alphabetical order as they are for `Fedora`, it would make it easier to compare the lists.  Fixing the `Ubuntu` list in the same way would make everything consistent.

eLtMosen:
> If the packages were listed in alphabetical order as they are for `Fedora`, it would make it easier to compare the lists. Fixing the `Ubuntu` list in the same way would make everything consistent.

Good idea, done!

beroset (approved):
Looks good.  Thanks!

PR #207, opened 2022-12-29
watches: Update `beluga` with the most recent updates.

MagneFire:
Additionally, mark USB mode as partial as there are some issues with getting it to switch to developer mode.

Only merge this once https://github.com/AsteroidOS/meta-smartwatch/pull/141 is merged.


FlorentRevest (commented):
Maybe this watch should get an extra star now ?

MagneFire:
I've been thinking about that too but there are a couple of reasons why I didn't do this yet:
- Performance isn't still on-par with the other Snapdragon 400 based watches.
  - Even `catfish` appears to be smoother which is a Wear 2100  based watch, while `beluga` is a Wear 3100 based watch.
- The installation requires you to flash the `boot` or `recovery` partition. Making a true temp install not possible as `fastboot boot` doesn't actually work.
- AsteroidOS isn't optimized for non-square/circular displays. There are various places that don't scale well on `beluga`
  - Charging rings are not circular.
  - Pairing prompts are off-center
  - Default launcher is off-center
  - Watchface 'bezels' are not circular but an oval shape.

PR #211, opened 2023-01-13
Add notes on creating and using SDK for emulator

beroset:
This adds details on how to get, build, install, and use the version of the SDK that targets qemux86 and therefore the emulator.


Lelaxlp:
Can you add the approximate size of the installation to the docs?
It would be a nice heads up, I think :)

pages/wiki/creating-an-asteroid-app.md
```diff
@@ -163,6 +168,29 @@ If you want to disable screen locking for easier development you can enable the
 ```
 mcetool -D on
 ```
+
+# Building applications for the emulator
+The primary difference between the two versions of the SDK are that the real watches use an ARM processor, which the emulator uses a simulated x86 processor.  This allows the emulator to run much faster on computers based on the x86 architecture than if they had to emulate an ARM processor.
```
FlorentRevest, line 173: s/which/while/

pages/wiki/creating-an-asteroid-app.md
```diff
@@ -163,6 +168,29 @@ If you want to disable screen locking for easier development you can enable the
...
+
+## Prebuilt emulator SDK
+You can either download a prebuilt emulator SDK from the nightlies [here](https://release.asteroidos.org/nightlies/sdk/oecore-x86_64-core2-32-toolchain-nodistro.0.sh) and install it on your system or you can build it yourself in a process very similar to that described above for creating the watch version of the SDK.
+
+## Building and installng the emulator SDK
```
FlorentRevest, line 178: s/installng/installing/

pages/wiki/creating-an-asteroid-app.md
```diff
@@ -163,6 +168,29 @@ If you want to disable screen locking for easier development you can enable the
...
+
+```
+./tmp-glibc/deploy/sdk/oecore-x86_64-core2-32-toolchain-nodistro.0.sh
+```
+
+## Using the emulator SDK
+As the emulator installer itself describes when it is first installed,
+
+>    Each time you wish to use the SDK in a new shell session, you need to source the environment setup script e.g.
+>     $ . /usr/local/oecore-x86_64/environment-setup-armv7vehf-neon-oe-linux-gnueabi
```
FlorentRevest, line 189: If we're talking about the emulator SDK maybe you shouldn't source the arm version ?

pages/wiki/creating-an-asteroid-app.md
```diff
@@ -163,6 +168,29 @@ If you want to disable screen locking for easier development you can enable the
...
+>     $ . /usr/local/oecore-x86_64/environment-setup-core2-32-oe-linux
+
+Integration with QtCreator is not explicitly described here, but again, the process is very similar to the process for useing QtCreator with the watch version of the SDK.
```
FlorentRevest, line 192: s/useing/using/

pages/wiki/creating-an-asteroid-app.md
beroset: Good catch.

pages/wiki/creating-an-asteroid-app.md
beroset: Good point.

beroset:
> Can you add the approximate size of the installation to the docs? It would be a nice heads up, I think :)

Done!  Thanks for the suggestion.  FYI, on my machine with multiple builds for `catfish`, `sturgeon`, `dory`, and `qemux86` the build directory takes about 177GiB.  Because it depends on what you've built, I just say in the docs that >150GiB is not unusual.

Lelaxlp:
@beroset 
https://release.asteroidos.org/nightlies/sdk/oecore-x86_64-core2-32-toolchain-nodistro.0.sh
returns a 404

beroset:
> @beroset https://release.asteroidos.org/nightlies/sdk/oecore-x86_64-core2-32-toolchain-nodistro.0.sh returns a 404

Yes, that's due to a known temporary issue with the server.  It's being worked on.

PR #212, opened 2023-01-13
Update emulator instructions

beroset:
This fixes a typo, adds some specifics relevant to newer versions of QEMU (that is, we now treat new versions as the norm and old versions as the exception) and removes a TODO note that points to a dead link.


jrtberlin:
Thank you!

PR #213, opened 2023-01-15
Fixed wrong link tag

wannaphong:
Hello! I found the features table link in https://asteroidos.org/install/ that It's not working because It has wrong tag.


MagneFire:
Great find!

PR #218, opened 2023-01-23
Change banner to png

eLtMosen:
@MagneFire noted that the banner has no transparency to work with dark mode.
This PR solves this while trying to keep the file size at a minimum by resizing it to 1140px width used on common Fullhd resolution.
The shadow is suboptimal in forced dark mode that alters colors like seen in below screenshot. Note also the asteroid logo being shifted towards lighter colors. Might try to fix that later.


eLtMosen:
grafik

MagneFire:
Thank you for your swift actions!

PR #219, opened 2023-01-23
Clarify bluetooth

beroset:
This clarifies a number of things in the Bluetooth LE profiles page.

1. it changes to MD format rather than HTML to make it simpler to edit
2. it corrects and expands the Time Service explanation
3. it corrects and expands the Weather Service explanation and adds examples

### DIFF
--- a/pages/wiki/ble-profiles.md
+++ b/pages/wiki/ble-profiles.md
@@ -3,84 +3,87 @@ title: Bluetooth Low Energy Profiles
 layout: documentation
 ---
 
-<p>In order to provide data synchronization capabilities between a watch and another device (typically a phone but that could also be a computer), AsteroidOS defines a variety of Bluetooth Low Energy profiles. This page tries to document those profiles so that developers can implement new synchronization apps.</p>
+In order to provide data synchronization capabilities between a watch and another device (typically a phone but that could also be a computer), AsteroidOS defines a variety of Bluetooth Low Energy profiles. This page tries to document those profiles so that developers can implement new synchronization apps.
 
-<p><a href="https://learn.adafruit.com/introduction-to-bluetooth-low-energy/gatt#services-and-characteristics">Basic knowledge of BLE</a> is useful to understand this page. In this page, the <i>server parts</i> refer to the side of the protocol implemented by <a href="https://github.com/AsteroidOS/asteroid-btsyncd">asteroid-btsyncd</a> on the watch. The <i>client parts</i> refer to what's implemented by external apps such as <a href="https://github.com/AsteroidOS/AsteroidOSSync/">AsteroidOSSync</a>.</p>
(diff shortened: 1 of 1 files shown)


pages/wiki/ble-profiles.md
```diff
@@ -3,140 +3,183 @@ title: Bluetooth Low Energy Profiles
...
+
+# Weather Profile
+
+A client can fill weather information buffers on the watch using the Weather Profile. Those information can typically be used by asteroid-weather or asteroid-launcher's Today page to show current info.  For each of the characteristics except for the Weather City Characteristic, the data is interpreted as a sequence of five two-byte big-endian values that correspond to "today" through "today + 4 days".
+
+### Weather Service (UUID: 00008071-0000-0000-0000-00A57E401D05)
+
+**Weather City Characteristic (UUID: 00008001-0000-0000-0000-00A57E401D05)**
+
+This characteristic can be written by the client to set the name of the city whose weather information are attached to. Values should be UTF-8 strings and they do not need to be NUL-terminated.
```
MagneFire, line 143: Typo here: This should be `NULL-terminated`

MagneFire (commented):
Overal this looks like good additional information as well as clearing up some confusion. Thanks for this!

Just a really minor nitpick on my side.

pages/wiki/ble-profiles.md
beroset: It's confusing, but `NULL` and `NUL` are two different things.  `NULL` is a null pointer, while `NUL` is the name of a character with the value of zero.  ASCII strings are often terminated with a `NUL` character, and I wanted to point out the fact that such a terminator is **NOT** required here.  I'm open to suggestions for better wording.

MagneFire: I did not know that, thanks for this!

Your example of ASCII strings doesn't seem to be consistent looking at my first search results (`NUL` and `NULL` appear to be mixed and matched). Regardless, it does show the difference very well.

pages/wiki/ble-profiles.md
```diff
@@ -3,140 +3,183 @@ title: Bluetooth Low Energy Profiles
...
+0x0111 = 273K = 0°C
+0x0211 = 529K = 256°C
+0x0311 = 785K = 512°C
+0xffff = 65535K = 65262°C
+0x0000 = 0K = -273°C
+```
+
+**Weather Max Temperatures Characteristic (UUID: 00008004-0000-0000-0000-00A57E401D05)**
+
+This characteristic can be written by the client with a value of 10 bytes. Each pair or byte represents a maximum temperature for a specific day of forecast. Interpretation is identical to that of the Weather Min Termperatures Characteristic.
```
FlorentRevest, line 185: Nit: temperatures* :)

FlorentRevest (approved):
Looks good to me! I just found a typo but nothing important

pages/wiki/ble-profiles.md
beroset: Thanks!  Also I changed "pair or byte" to "pair of bytes" in each occurrence.

Issue #193, opened 2022-10-06
Integrate model/product code and possibly more differentiating details for all watches


eLtMosen:
Currently, only some install pages have the product code of a watch mentioned in not so prominent place.
Product code however
- Help users identify the watch exactly when purchasing
- Free SEO regrading requests for product codes from search engines.
- Generally give more detailed info on the watches and act as trusted source in the matter.

Lets discuss where to put this specific info. And possibly how to extend it to other details that differ between models like strap width and casing size.

This just to start off the product code collection, please do not mind the formating at all and ideally past your corrections in a better formated list :sweat_smile: 

:heavy_check_mark: W100 = dory [LG G Watch]
:heavy_check_mark: W110 = lenok [LG G Watch R]
:heavy_check_mark: W150 = bass [LG G Watch Urbane]
:heavy_check_mark:LM-W315 = narwhal [LG Watch W7]

:heavy_check_mark: 360L = smelt [Moto 360 2nd gen, 46mm, 20mm]
:heavy_check_mark: 360S = carp [Moto 360 2nd gen, 42mm, 18mm]
:heavy_check_mark: 360W = carp [Moto 360 2nd gen, 42mm, 16mm]
:heavy_check_mark: 360SP = sportscarp [Moto 360 sports]

:heavy_check_mark: WF11018 = catfish-ext [TicWatch Pro LTE]
:heavy_check_mark: WF12096 = catfish [TicWatch Pro]
:heavy_check_mark: WF12106 = catshark [TicWatch Pro 2020]
:heavy_check_mark: WG12016 = tunny [TicWatch S2]
:heavy_check_mark: WG12036 = skipjack [TicWatch C2/C2+]
:heavy_check_mark: WF12086 = mooneye [TicWatch E]
:heavy_check_mark: WF12066 = mooneye [TicWatch S]

:heavy_check_mark: WI500Q = anthias [Asus ZenWatch]
:heavy_check_mark: WI501Q = sparrow [Asus ZenWatch 2, 46mm]
:heavy_check_mark: WI502Q = wren [Asus ZenWatch 2, 41mm]
:heavy_check_mark: WI503Q = swift [Asus ZenWatch 3]

:heavy_check_mark: LEO-BX9 = Sawfish [Huawei Watch 2]
:heavy_check_mark: LEO-DLXX = Sawshark [Huawei Watch 2 LTE]

:heavy_check_mark: DW6F1 = firefish [Fossil Gen4]
[...]

PR #222, opened 2023-02-13
[Install pages ]Add all known model numbers and polish install pages

eLtMosen:
This PR aims to integrate the model numbers for most of the watches to their install pages, like noted in this issue:
https://github.com/AsteroidOS/asteroidos.org/issues/193
All model numbers have been researched to be correct. Multiple sources per search for each code where easily verifiable.
The advantage of listing the model numbers are:
- Users can better identify and verify the system image for the watch they are owning.
- Purchases on second hand market often yield much better results when adding the model number.
- Search engines will index the model numbers and list us in searches accordingly.

This PR also fixes the narwhal wlan feature matrix to "bad".
And adds E2 tunny to the install main page and feature matrix name for skipjack.

Currently model numbers for following watches are missing:
- minnow [moto 360 gen 1]
- sturgeon [Huawei Watch]
- harmony/inharmony

I decided to add a blue infobox as wrapper. Thanks to @MagneFire and @dodoradio for giving valuable input on the layout.
The current approach only repeats watch names when necessary to differentiate variants. For watches with only one variant, only the model number is shown.
Compare a complex case in firefish/ray:
grafik

or OPPO Watch
grafik

With e.g. narwhal:
grafik

### DIFF
--- a/pages/install/anthias.md
+++ b/pages/install/anthias.md
@@ -3,3 +3,7 @@ title: Asus Zenwatch 1
 deviceName: anthias
 layout: aw-install
 ---
+<div class="callout callout-info">
+    <h4>Model number</h4>
+    <p>WI500Q</p>
+</div>
--- a/pages/install/bass.md
+++ b/pages/install/bass.md
@@ -4,6 +4,10 @@ deviceName: bass
 section: install
 layout: aw-install
 ---
+<div class="callout callout-info">
+    <h4>Model number</h4>
+    <p>W150</p>
+</div>
 
 <div class="callout callout-info">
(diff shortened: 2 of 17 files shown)


FlorentRevest:
Neat! :) I think it would ultimately be cool to centralize all of these fields into `pages/watches-support.json` and have the install page template generate the callout div using infos from there but that's already a good step in the right direction :)

eLtMosen:
Agreed! Foremost wanted to find a format how to publish them and document them properly for the time being.
Thid was sort of a personal 2.0 blocker :P

PR #224, opened 2023-02-15
Use templates for model info

beroset:
This make some changes to the styles and layouts to allow for generating the watch model number tables using templates rather than hardcoding.  This should make it easier to maintain and augment in the future, since all the information is now coming from a single json file.

Three things I'd like particular attention to: 

1. The firefish/ray wording was changed a bit to make it make more sense in this new context, but I want to make sure we preserve the excellent work of @eLtMosen so I would ask that reviewers pay particular attention to that.

2. The pale yellow color I've chosen for the table highlights looks good to me, but I'd like other's opinions on whether some other value would be better.  In particular it's similar to but not identical to the color of warning text (as with `minnow`) so maybe another color would be better.

3. The list of watches for `harmony` is long and I'm not sure about all of the models.  For instance the Lemfo LES1 is listed under both `harmony` and `inharmony`, as is the Diggro DI07 (but with "older" specified for `inharmony`) and I couldn't find any information about the IQ I2 watch.   Those are not changes, but I noticed them when I was reviewing the pages.


beroset:
Changed the table row highlight color to light blue per suggestion from @eLtMosen

beroset:
Here's what it currently looks like:
image

templates/layouts/aw-install-multi.hbs
```diff
@@ -31,6 +31,19 @@
         {{/with}}
         </table>
       {{/deviceNames}}
+      </table>
```
eLtMosen, line 34: I would suggest to put at least one spacing-row between the two tables. The hardware feature headings have various bottom spacings to the info table. Due to tilt-to-wake wrapping into 3 lines.
With spacing between the tables, the new model headings would be more obvious separated from the colored feature table.

eLtMosen (changes requested):
Just the spacing between the old and new table. Else its great imho!

templates/layouts/aw-install-multi.hbs
beroset: Good idea.  It does look better.
image

eLtMosen (approved):

eLtMosen:
Ow, i just noted a typo not related to your PR but possibly worth to fix while we are at it.
There are instances where `Kingwear` is written `Kingswear` falsly.
Do you see those and could quickly correct? Else i'd do a follow up PR.

PR #225, opened 2023-02-17
Add info fix typo

eLtMosen:
Found an italian source with all DW6xx and DW7xx models listed. Updated the firefish/ray model numbers.
https://orologismartwatch.com/fossil-dw6-platform
https://orologismartwatch.com/fossil-dw7-platform

Also found that W1 is used a lot for sturgeon.


MagneFire:
Looks good to me! Thanks for digging up this information!

PR #231, opened 2023-02-21
Remove ie9 support

beroset:
Since IE9 was EOL'd three years ago, we can simplify by dropping support for that antique browser.


FlorentRevest:
Wait what ? I'm still using IE9, what's wrong with it ?

PR #234, opened 2023-02-22
Add kernel versions

beroset:
This adds "soc", "kernelversion" and "androidversion" as three new fields in the json file.  The data was populated using some awk scripts against the meta-smartwatch git repository.  The data is not yet used, but will be useful for an upcoming PR to show hardware and software details for each watch.


FlorentRevest:
Please don't tell anyone how old our kernels are

PR #236, opened 2023-02-22
Shorten headlines to make the menubar better legible and not clip

eLtMosen:
image

This solves the clipping

image

But question is if the text is possibly too short now.


beroset (approved):
Nice!  I think this is an improvement and makes it easier to scan the questions in the sidebar.

eLtMosen:
Visual after furtheer optimization

image

FlorentRevest:
Thanks, I like this sort of simplification. Generally speaking our website should be "straight to the point". That's a good step already

FlorentRevest:
(although this clipping is a bug and should be fixed separately)

PR #241, opened 2023-02-24
Add note about generic iOS BLE apps

GeckoEidechse:
In particular how they can be used to connect a watch running AsteroidOS to an iOS device.

I came across [nRF Connect](https://apps.apple.com/gb/app/nrf-connect-for-mobile/id1054362403) while checking the [PineTime wiki](https://wiki.pine64.org/wiki/PineTime_FAQ#How_do_I_set_the_time_on_PineTime?) and found it also works with my lenok running AsteroidOS. I don't have other watches so I cannot confirm whether it works on all devices running AsteroidOS.

This is also by no means a replacement for a proper sync client but certainly better than no sync at all :P

(Now if someone could just fork [InfiniLink](https://github.com/InfiniTimeOrg/InfiniLink) and adapt it for AsteroidOS so that we have a proper sync client on iOS, that would be great x_x)

### DIFF
--- a/pages/wiki/synchronization-clients.md
+++ b/pages/wiki/synchronization-clients.md
@@ -24,7 +24,7 @@ layout: documentation
 
 <p>There is currently no application available for iOS phones to synchronize with AsteroidOS watches. As far as we know, there hasn't been any development done on such an app yet.</p>
 
-<p>One can however use generic BLE scanners/communication apps like <a href="https://apps.apple.com/gb/app/nrf-connect-for-mobile/id1054362403">nRF Connect</a> to still connect the watch to an iOS device and even use it to send notifications to the watch. Specific functionality like syncing time or weather forecast are however not supported!</p>
+<p>However, if the watch is paired with an iOS device, AsteroidOS is capable of displaying notifications from the phone. Pairing can be done using Apple's Bluetooth settings, but several iOS users have reported that a generic BLE scanner/communication app like <a href="https://apps.apple.com/gb/app/nrf-connect-for-mobile/id1054362403">nRF Connect</a> seems to be more reliable for pairing Bluetooth devices. Specific functionality like syncing time or weather forecast are however not supported!</p>
 
 <div class="page-header">
   <h1 id="sfos">SailfishOS application</h1>
--- a/pages/wiki/synchronization-clients.md
+++ b/pages/wiki/synchronization-clients.md
@@ -24,6 +24,8 @@ layout: documentation
 
 <p>There is currently no application available for iOS phones to synchronize with AsteroidOS watches. As far as we know, there hasn't been any development done on such an app yet.</p>
 
+<p>One can however use generic BLE scanners/communication apps like <a href="https://apps.apple.com/gb/app/nrf-connect-for-mobile/id1054362403">nRF Connect</a> to still connect the watch to an iOS device and even use it to send notifications to the watch. Specific functionality like syncing time or weather forecast are however not supported!</p>
+
 <div class="page-header">
   <h1 id="sfos">SailfishOS application</h1>
 </div>


pages/wiki/synchronization-clients.md
```diff
@@ -24,6 +24,8 @@ layout: documentation
 
 <p>There is currently no application available for iOS phones to synchronize with AsteroidOS watches. As far as we know, there hasn't been any development done on such an app yet.</p>
 
+<p>One can however use generic BLE scanners/communication apps like <a href="https://apps.apple.com/gb/app/nrf-connect-for-mobile/id1054362403">nRF Connect</a> to still connect the watch to an iOS device and even use it to send notifications to the watch. Specific functionality like syncing time or weather forecast are however not supported!</p>
```
beroset, line 27: Like you, I find that nRF Connect works better, but one can pair the watch with just the Apple settings.  Given that, I'd suggest that we might change it to say something like:

"However, if the watch is paired with an iOS phone, AsteroidOS is capable of displaying notifications from the phone.  Pairing can be done using Apples Bluetooth settings, but several iOS users have reported that a generic BLE scanner/communication app like {nRF connect with hyperlink} seems to be more reliable for pairing Bluetooth devices.  Specific functionality ...."

beroset (changes requested):
Thanks for the contribution.  I have one suggested change.

pages/wiki/synchronization-clients.md
GeckoEidechse: Done in 68e82e7433c19172eff08d67610e3020eaa537b2

beroset (approved):
Good contribution.  Thanks!

FlorentRevest:
It feels to me like this paragraph lets users believe they will have notifications sync with nRF Connect. I'd rather describe nRF Connect as a debug or development tool to interact with the AsteroidOS BLE profiles at a low level. Sending notifications through it doesn't sound very useful otherwise 😅

beroset:
I think the revised paragraph takes care of that by explicitly stating that:

> Pairing can be done using Apple's Bluetooth settings, but several iOS users have reported that a generic BLE scanner/communication app like <a href="https://apps.apple.com/gb/app/nrf-connect-for-mobile/id1054362403">nRF Connect</a> seems to be more reliable for pairing Bluetooth devices.

If not, please suggest alternative wording.

GeckoEidechse:
> It feels to me like this paragraph lets users believe they will have notifications sync with nRF Connect. 

Not sure what your exact definition of _notification sync_ is, so to clarify, when pairing my iPhone to my lenok via _nRF Connect_, any push notification that would arrive on my iPhone would also show up on my lenok.

Dismissing the notification on either however does not dismiss the notification on the other. So the "sync" is kinda one-way :P

However it certainly does more than just allowing you to send test notifications from the app to the watch. Actually I'm not even sure the app allows doing that ^^"

pages/wiki/synchronization-clients.md
GeckoEidechse: Based om what _DocGalaxyBlock_ said in the Matrix channel ([archive link](https://log.asteroidos.org/?date=2023-02-24&level=16#184)), it seems that being able to connect an iDevice to a watch running AsteroidOS via standard Bluetooth settings seems to be the exception, rather than the norm.

They tested `sturgeon`, `sparrow` and `narwhal` on a MacBook.

image

beroset: That's a separate issue.  This paragraph talks about iOS, not OSX.

FlorentRevest:
> I think the revised paragraph takes care of that by explicitly 

Oh, sorry, for some reason the diff I was shown by github yesterday was the previous one. The paragraph I get to read today seems fine indeed

> However it certainly does more than just allowing you to send test notifications from the app to the watch.

And that is also true given that we support ANCS indeed. Somehow, I was focused on my own usage of nRF Connect (for debugging) and must have been too tired to remember the iOS context...

Neeeveeermind. Sorry about that, merging now. Thanks for the contrib!

PR #242, opened 2023-02-25
Merge sparrow wren

eLtMosen:
Remove wren from install page

image

And create aw-install-dual.hbs to work with dual system image but one install item watches

image


FlorentRevest:
One day we should consolidate all these layout files, they are meant to avoid duplication but we duplicate the layouts ahah...

Anyway, today is not that day

PR #245, opened 2023-02-25
Rename firefish and ray to Fossil Gen 4 46mm and 41mm

eLtMosen:
Due to recent addition of many more firefish and ray models, the skagen falster 2 name for ray became unfitting.
Renaming to the main differentiationg factor, the casing size.

image

### DIFF
--- a/pages/watches-support.json
+++ b/pages/watches-support.json
@@ -156,7 +156,7 @@
     {
       "name":"firefish",
       "models": [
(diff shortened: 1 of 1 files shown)


eLtMosen:
Screenshot has "Fossil Gen4 ..." wrong. Is corrected to "Fossil Gen 4 ..." via force push.

PR #248, opened 2023-02-26
Add crown to feature matrix and feature table of supported watches

eLtMosen:
As brought up by @docgalaxyblock and discussed with @MagneFire 
This PR adds "Crown" as a feature to watches that got the hardware to make the feature more easily discoverable and document it properly.

image

image

image


FlorentRevest:
Makes me realize we should get rid of these empty parenthesis in the title in your first screenshot.

"Feature list ()"

PR #252, opened 2023-03-02
Make features table responsive

eLtMosen:
The features table did wrap unfortunate on mobile devices.
This PR
- Makes the support-column width depend on viewport width
- Limits the minimum columns size to 15px where it was 25px before
- Removes the support-row wrap, introducing horizontal scroll on width lower than 600px.
- Sets a max-width of 1024px for the support-row so it does not get wider than the top menu

Downsides 
- The columns are rectangles, not squares anymore.
- This is hardcoded to 3.5VW and needs to be lowered in case more columns get added

Before, reported by @thorgalix21 on matrix:
grafik

Behaviour with PR implemented:

https://user-images.githubusercontent.com/15074193/222314645-da10b5e6-f40d-431d-8ced-265b75f08f79.mp4

### DIFF
--- a/styles/components/install-support-table.less
+++ b/styles/components/install-support-table.less
@@ -7,11 +7,11 @@
   display: -webkit-flex;
   display: -ms-flexbox;
   display: flex;
-  flex-wrap: wrap;
+  max-width: 1024px;
 }
 
 th.rotated {
-    width: 30px;
+    width: 3.5VW;
     height: 40px;
     white-space: nowrap;
     transform: translateX(10px) rotate(-45deg);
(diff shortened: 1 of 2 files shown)


FlorentRevest:
While we are at it maybe it's also time to remove these parentheses in the title ?

This section is a install-header and the logic to show parenthesis is https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/includes/install-header.hbs#L8 what happens if you replace the else with an else if deviceName ?

eLtMosen:
> While we are at it maybe it's also time to remove these parentheses in the title ?
> 
> This section is a install-header and the logic to show parenthesis is https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/includes/install-header.hbs#L8 what happens if you replace the else with an else if deviceName ?

it was 
`{{#if deviceName}}({{deviceName}}){{else if label}}({{label}}){{/if}}`
Else the parenthesis would be outside of the if/else logic and always show.

FlorentRevest:
Ah yes I must have misread, but great, thank you for the fix!

Issue #203, opened 2022-11-18
Replace ClockworkMod UniversalAdbDriver with official google issued Android Platform-Tools ADB driver


eLtMosen:
Two users now reported the Windows ADB driver we link too does not work anymore. For one user, the official google platform-tools driver worked. The other case still in testing as i am writing this.

Problems is, where to find a linkable source for a compiled version of the driver.
Thats not from some shady google docs folder like this one:
https://drive.google.com/drive/folders/1H8DxaRta8gafcUA8bJqkajobnnfuw2et

E-Gwen:
Hi,

Unforntunately, both 32 & 64 bit adb drivers won't recognized the Asus Zen Watch 2 sparrow model I try to flash with AsteroidOS.
So it seems I can't go further at this time.
Ouf of ideas.

E-Gwen:
Hey, wait a minute 👍 
I downloaded the Asus adb driver here, https://drivers.softpedia.com/get/MOBILES/ASUS/ASUS-Android-ADB-Interface-Driver-8003-for-Windows-10.shtml, installed them,
and now I can access the menu to unlock the bootloader.

But when i wrote the flash image line, i get this error message : 
"cannot determine image filename for 'C:\Users\Gwen\asteroid-image-sparrow.ext4"

Can you please tell me what I'm doing wrong ?

Thanks.

E-Gwen:
Hey, wait a moment !
I forgot "userdata" in the command line "fastboot flash userdata %systemdrive%%homepath%\Downloads\asteroid-image-*.ext4"

Be careful, Windows newbies users !

I made it !

Can't wait the AsteroidOS 2.0 version !

Have a nice watch ! See you !

PR #258, opened 2023-03-03
Remove Clockwork mod USB driver and add Google USB driver

eLtMosen:
It was noted in the matrix chat that the download source seemed a bit questionable.
In addition, some users reported that it is not working with latest windows versions.

- Link to the Google Developer page containing a ZIP at https://developer.android.com/studio/run/win-usb
- Direct link to the file is not possible
- Open link in new tab/window (browser setting) to not loose the install page scroll state
- Announce the new tab/window using the `title` feature to show a hint on mouse hover.

This fixes https://github.com/AsteroidOS/asteroidos.org/issues/203

Visually:
grafik

### DIFF
--- a/templates/layouts/aw-install-dual.hbs
+++ b/templates/layouts/aw-install-dual.hbs
@@ -61,9 +61,9 @@
         <br><br>
         <pre><code class="bash">sudo apt install android-tools-adb android-tools-fastboot</code></pre>
         <br>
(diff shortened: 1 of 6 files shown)


FlorentRevest:
Ah that's a much needed improvement, I was also looking forward to that! Thanks :D

PR #261, opened 2023-03-05
Rename Install page to Watches page

eLtMosen:
Following a discussion on matrix, this PR renames the install page and takes care of all dependencies and links to it.
The install page always had the character of more than just being a referral to the install pages. But serving as sort of a showcase of all our supported watches. And often being an entry point for users looking to purchase a watch
Also in light of discussion to build a detailed watch wiki, this change aims to prepare for that addition.

grafik


FlorentRevest:
Ooc, for backward compatibility, what happens if you create a "install" symlink to "watches" now ? So that if people linked to our current install pages, the links would work. (at least a copy, at best a redirection to the watches pages)

eLtMosen:
> Ooc, for backward compatibility, what happens if you create a "install" symlink to "watches" now ? So that if people linked to our current install pages, the links would work. (at least a copy, at best a redirection to the watches pages)

I circled around myself in matrix chat a bit and came to the conclusion:
Since we are using nginx, a clean 301 redirect that helps web crawlers to update their index to the new page is done in the nginx config like so
```
server {
  # …
  location = /install {
    return 301 watches
  } 
  # …
}
```
How can we implement this?

eLtMosen:
Conflicts resolved! I guess we just need to sync the merge with a config addition to nginx for the 301?

FlorentRevest:
Thanks! @jrtberlin you can do your magic :)

jrtberlin:
Done

Issue #263, opened 2023-03-05
[FAQ] Reorder <h1> use and optimize FAQ title


eLtMosen:
@FlorentRevest via matrix:
we just have a h1 "faq" at the top which isn't useful given that there's already a "Frequently asked question" in the breadcrumb. and when the page is open, the left navigation bar only contains that h1 which makes it look a bit awkward. but as soon as one scrolls a little it, all the questions (h2) appear in the side bar. I propose we just make every question a h1 and drop that "FAQ" top level title in the document

PR #270, opened 2023-03-06
Drop Header and change all markdown h2 to h1

eLtMosen:
Fixes https://github.com/AsteroidOS/asteroidos.org/issues/263

On page load now:
grafik

### DIFF
--- a/pages/main/faq.md
+++ b/pages/main/faq.md
@@ -1,7 +1,7 @@
 ---
 title: Frequently Asked Questions (FAQ)
 slug: faq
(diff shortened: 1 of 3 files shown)


eLtMosen:
Ow, i only now see the documentation slug is wrong.
Also, the edit pen is not working due to the faq.md page having moved away from the wiki subfolder.
Making an issue for the later https://github.com/AsteroidOS/asteroidos.org/issues/271
And fixing the slug in this PR.

Issue #273, opened 2023-03-06
[bug] mtk-install.md has a preperation headline doubled in wrong place


eLtMosen:
Just a note, it seems that you introduced a stray "Preparation" title before "Choose your variant" in the MTK6580 watches page https://asteroidos.org/install/harmony/

_Originally posted by @FlorentRevest in https://github.com/AsteroidOS/asteroidos.org/issues/253#issuecomment-1455176510_

PR #275, opened 2023-03-07
Remove stray Preparation header

eLtMosen:


eLtMosen:
Ow, this fixed https://github.com/AsteroidOS/asteroidos.org/issues/273

PR #276, opened 2023-03-07
Redesign Documentation User/Developer menu into 2 columns

eLtMosen:
@FlorentRevest wish was to see how the documentation menu would look in 2 columns.
Here it is

https://user-images.githubusercontent.com/15074193/223292174-42f9482f-1dc5-4046-a6c9-132013d35fca.mp4


FlorentRevest:
Great! This is already an excellent incremental change I'm very happy to take.

But it feels like the sidebar is useless now. I think we should modify the documentation template here https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/layouts/documentation.hbs#L17 to make the sidebar's existence conditional on the slug https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/layouts/content.hbs#L10  provided here https://github.com/AsteroidOS/asteroidos.org/blob/master/pages/wiki/documentation.md?plain=1#L3

And generally speaking, _at some point_ we should factorize a bunch of our templates using this sort of conditionals.

PR #279, opened 2023-03-07
Add nightstand news item

beroset:
This is a new news article that describes a little of what Nightstand mode is and how it came about.

PR #260, opened 2023-03-03
News post about gadgetbridge support

eLtMosen:
Text might need some rewording and could use some extension.
But @dodoradio mentioned to have a look into that <3

grafik

PR #280, opened 2023-03-08
Add markdown for news items

beroset:
This makes a minor adaptation to the news template to allow authors to write news items using markdown syntax.  This may help #260 and #279


FlorentRevest:
That's great! Thanks for the swift action :D

Issue #271, opened 2023-03-06
[FAQ] Edit pen (link to github) not correct


eLtMosen:
FAQ has moved from the wiki folder.  We need to either adapt or duplicate `/templates/includes/popover-source-link.hbs` where the wiki part of the link is currently hardcoded
https://github.com/AsteroidOS/asteroidos.org/blob/38546b64217751cddcc5fff6c207cf8cc8251fcb/templates/includes/popover-source-link.hbs#L7

PR #281, opened 2023-03-09
Fix edit pen github link

beroset:
This fixes issue #271 by correctly constructing the link to the github source page no matter where in the hierarchy the page exists.


eLtMosen:
Yepp, double checked and works! Thank you.

PR #282, opened 2023-03-09
Rework old news posts images

eLtMosen:
The images where supplied in higher resolution than ever actually used.
- Resize images to maximum size used 
- Convert to webp
- Thus reduce size by factor of 10.


eLtMosen:
Ugh, what happened here? Closing...

PR #285, opened 2023-03-12
Redesign ADB unlock screenshots and reorder headline styles

eLtMosen:
This PR aims at unifying the look of the install pages after adding the new install feature boxes.

- Split adb unlock images to individual screenshot so they look consistent with the planned startpage slider.
- Make the screenshots float left so they wrap correctly on smaller display sizes
- Change highlight boxes to consistent size to suggest finger size during touch 
- Add rounded edges to highlight boxes to match the install feature boxes style
- Change to green from red to appear more friendly 
- Adapt to green highlight color to match the new install feature boxes good color
- Resize all images to 240px maximum width to reduce file size
- Unify and reorder h1, h2, h3 and h4 hierarchy

This is how the page behaves after the PR:

https://user-images.githubusercontent.com/15074193/224579652-0f5c2d00-f68f-48e7-b1e7-808f7a83ca6e.mp4

two column order stays up to 375px phone width

grafik

three columns on tablets
grafik

And beluga
grafik


templates/layouts/beluga-install.hbs
```diff
@@ -66,34 +66,55 @@
...
+          <img src="{{assets}}/img/install-beluga-01.webp" class="unlock-step-img">
+          <div class="unlock-step-text">5. Open the <b>Settings</b> app again</div>
+        </div>
+        <div class="unlock-step-wrapper">
+          <img src="{{assets}}/img/install-beluga-05.webp" class="unlock-step-img">
+          <div class="unlock-step-text">6. Open the <b>Developer options</b> page</div>
+        </div>
+        <div class="unlock-step-wrapper">
+          <img src="{{assets}}/img/install-beluga-06.webp" class="unlock-step-img">
+          <div class="unlock-step-text">7. Enable <b>ADB debugging</b> and <b>OEM unlock</b></div>
```
FlorentRevest, line 102: Iiuc, this unlocks the bootloader right ?

Is step 1.2 necessary then ?

templates/layouts/mooneye-install.hbs
```diff
@@ -66,34 +66,55 @@
...
-        </ol>
-      <br><b>Finally, to unlock the bootloader</b> enter the following commands in a terminal with your watch connected:</li>
+      <h2>1. Unlock your bootloader</h2>
+      Installing AsteroidOS requires an unlocked bootloader.
+      <br>Accessing the fastboot bootloader menu on your watch can be achieved via ADB in Wear OS, as described below in step 1.1.
+      <br>You can skip to step 1.2 if you are familiar with the manual method to access the bootloader menu. The methods are explained on the <a href="https://asteroidos.org/wiki/useful-commands/#boot-to-fastboot-bootloader-menu">useful commands page</a>.
+      <h3>1.1 Enable ADB on your watch with the following steps:</h3>
+      <div class="unlock-adb-wrapper">
+        <div class="unlock-step-wrapper">
+          <img src="{{assets}}/img/install-aw-01.webp" class="unlock-step-img">
```
FlorentRevest, line 77: For some reason this one image loads as a grey square for me, do you have the same issue ? https://raw.githubusercontent.com/AsteroidOS/asteroidos.org/e00728d931cd2c3c28fda35df89931141ff49897/assets/public/img/install-aw-01.webp

Other images work

FlorentRevest (commented):
Exciting :)

FlorentRevest:
Also, a cheap trick to avoid all that copy-pasting (and mistakes risk, remember we had a lot of h3/h2 inconsistency issues in the last PR touching these install pages) could be to extract the wearos bootloader unlock instructions into an include block, a bit like we do for the install problem section: https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/includes/install-problems.hbs included like this https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/layouts/aw-install-dual.hbs#L124

This could be a good first step towards consolidating all these templates https://github.com/AsteroidOS/asteroidos.org/tree/master/templates/layouts which are out of control already.

Alternatively, I've been talking a lot about using conditionals like https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/includes/documentation-header.hbs#L5 to make templates more dynamic based on parameters (here the condition is on the slug variable: https://github.com/AsteroidOS/asteroidos.org/blob/master/pages/wiki/documentation.md?plain=1#L3 but we could use arbitrary variables. For example, we could have

```
{{#is codename 'beluga'}}
  {{> bootloader-unlock-beluga}}
{{else}}
  {{> aw-bootloader-unlock-beluga }}
{{/is}}
```

Or

```
{{#is watch-needs-simg 'yes'}}
        <a class="btn btn-primary" href="https://release.asteroidos.org/nightlies/{{deviceName}}/asteroid-image-{{deviceName}}.simg" role="button">asteroid-image-{{deviceName}}.simg</a>
{{/is}}
```

Or `{{#is two-columns 'yes'}}` (for firefish/ray) etc... There is a balance to be found between using different templates or include blocks. For example, maybe the mtk-install is so different from the aw-install that they shouldn't be folded together, but maybe they should, I'm not sure about that one. But ultimately the goal should just be to 1- avoid duplication 2- Keep things modular and easy.

For example, one thing I find quite bizarre looking back on our code (I should have noticed it at the time of the pull request review :) ) is our empty `watches.md` https://raw.githubusercontent.com/AsteroidOS/asteroidos.org/master/pages/main/watches.md I would expect the content of that page to be in `watches.md` because this is just an empty `body` that gets included in https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/layouts/watches.hbs#L33 (weirdly, just after the content of the page is defined in the layout...)

Similarly, we could hide the navbar of the documentation landing page only when `{{#is slug 'documentation'}}`.

I've been restraining from being too annoying with this but I really think we're at the point where this sort of consolidation is needed to keep the website easy to evolve uh.

templates/layouts/beluga-install.hbs
eLtMosen: The OEM unlock switch is needed to allow unlock. `fastboot oem unlock` stilll needs to be issued but would fail if the option is not set.

templates/layouts/mooneye-install.hbs
eLtMosen: Yeah, thanks for confirming. I had a similar occurrence, but in my case the 1st and 5th image would not show on firefox at some point. I did ignore the issue since just opening the same state in any other brower then showed all images correctly.
We might just move back to png since this smells like a webp problem?...

eLtMosen: Ok, i guess i found the issue. That first picture was the only one i exported to "lossless" webp which seems to be not perfectly supported by all browsers. Now exported it to 90% image quality like the other and the problem disappeared for me.

eLtMosen:
> Also, a cheap trick to avoid all that copy-pasting (and mistakes risk, remember we had a lot of h3/h2 inconsistency issues in the last PR touching these install pages) could be to extract the wearos bootloader unlock instructions into an include block, a bit like we do for the install problem section: https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/includes/install-problems.hbs included like this https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/layouts/aw-install-dual.hbs#L124
> 

I have implemented new include boxes for round and beluga watches. The idea is to add square watches later. But for that we need an entry in the json wether a watch has round or square screen.

Extending the handle bars and all other conditional changes imho should not be part of this PR.
There are several complications, like beluga not only needing different unlock instructions but the flashing instructions differ to all watches additionally.
Also the simg conditional would hook into the install instructions which i planned to tackle in another PR. To implement that only the relevant install instructions are shown, based on dropdown selection.

I am going to make a new issue from your plans so we can work on that next.
https://github.com/AsteroidOS/asteroidos.org/issues/292

I can solve the current issues using github UI. But before doing so i would like to get this state approved since its a mess for me still to later commit to this branch after having solved the conflicts using the web ui.

templates/layouts/beluga-install.hbs
FlorentRevest: Ah I see, thank you

templates/layouts/mooneye-install.hbs
FlorentRevest: Neat, indeed the new picture works for me too.

FlorentRevest (approved):

FlorentRevest:
> > Also, a cheap trick to avoid all that copy-pasting (and mistakes risk, remember we had a lot of h3/h2 inconsistency issues in the last PR touching these install pages) could be to extract the wearos bootloader unlock instructions into an include block, a bit like we do for the install problem section: https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/includes/install-problems.hbs included like this https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/layouts/aw-install-dual.hbs#L124
> 
> I have implemented new include boxes for round and beluga watches.

Awesome! :) The reason why I insisted is because I wanted you to try it once just to "demistify" it. I think it's a good practice because it makes your life easier, my life as a reviewer easier and the life of future contributors easier too. :) 

> The idea is to add square watches later. But for that we need an entry in the json wether a watch has round or square screen.

Ahah, so much attention to detail. I trust users will not be too lost either if their watch is square and they see circular screenshots. But sure it's a nice addition. 

> Extending the handle bars and all other conditional changes imho should not be part of this PR. There are several complications, like beluga not only needing different unlock instructions but the flashing instructions differ to all watches additionally. Also the simg conditional would hook into the install instructions which i planned to tackle in another PR. To implement that only the relevant install instructions are shown, based on dropdown selection.

Agreed. I was just thinking out loud.

> I am going to make a new issue from your plans so we can work on that next. #292

🙏

> I can solve the current conflicts using github UI. But before doing so i would like to get this state approved since its a mess for me still to later commit to this branch after having solved the conflicts using the web ui.

Approved ;)

PR #286, opened 2023-03-14
Add checksum info

### LINKED ISSUE #284: Provide the `sha256sum` information on the install pages
Great idea! This makes me wonder if we should provide the `sha256sum` information on the install pages themselves too since we already know the checksums in advance: https://release.asteroidos.org/nightlies/beluga/MD5SUMS

(This also means that the `We do not currently provide checksums for the files.` claim, is incorrect :wink: )

_Originally posted by @MagneFire in https://github.com/AsteroidOS/asteroidos.org/pull/283#discussion_r1133270344_

beroset:
This fixes #284 by adding buttons for downloading the `SHA512SUMS` files for each watch and also amending the troubleshooting instructions to show how to use it.


templates/includes/install-problems.hbs
```diff
@@ -11,7 +11,7 @@
...
-      <li><h4>Repeat the flash process</h4>Some users have reported needing up to five flashing attempts for a successful result.<br>There is a small chance that the image files were corrupted during download. We do not currently provide checksums for the files. However, if you download the files again under a different filename and then compare them with <code>sha256sum filename</code>, the checksums must match.</li>
+      <li><h4>Repeat the flash process</h4>Some users have reported needing up to five flashing attempts for a successful result.<br>There is a small chance that the image files were corrupted during download. To guard against this, it's recommended to verify the checksums.  Download the <code>SHA512SUMS</code> checksum file from the same directory as the image files and verify them with <code>sha512sum --check SHA512SUMS</code>. The checksums for both files will be checked.</li>
```
eLtMosen, line 14: Very nice of you to promptly change the information text too!

templates/layouts/mtk-install.hbs
```diff
@@ -43,6 +43,7 @@
         <p>
           <a class="btn btn-primary" href="https://release.asteroidos.org/nightlies/{{.}}/asteroid-image-{{.}}.ext4" role="button">asteroid-image-{{.}}.ext4</a>
           <a class="btn btn-primary" href="https://release.asteroidos.org/nightlies/{{.}}/zImage-dtb-{{.}}.fastboot" role="button">zImage-dtb-{{.}}.fastboot</a>
+          <a class="btn btn-primary" href="https://release.asteroidos.org/nightlies/{{.}}/SHA512SUMS" role="button">SHA512SUMS</a>
```
eLtMosen, line 46: ehehe, i was actually wondering if might be that easy. Thanks for implementing so very swiftly.

eLtMosen (approved):
Tested and approved!
Very happy here about the momentum the website got currently :partying_face:

templates/includes/install-problems.hbs
FlorentRevest, line 14: What would you think about having

```
"Download the  <a href="https://release.asteroidos.org/nightlies/{{.}}/SHA512SUMS">SHA512SUMS</a><code>SHA512SUMS</code></a> file and verify them with...
```

Here and not introducing a button for hashes at the top of the install page.

I personally find that checksums are only really worth checking in case of troubleshooting (btw have you *ever* downloaded a file with an incorrect checksum over http ?) and therefore it seems we can avoid one unnecessary step in the "critical user journey" of installing the OS and only keep the link at the time when it's actually relevant.

Note I removed "from the same directory as the image files" from your text as the user isn't really expected to know these files are in the same directory (they can just click the link)

beroset: Done.  It's a bit more complex because some pages have multiple watches but this takes care of all that using templates.

templates/includes/install-problems.hbs
```diff
@@ -11,7 +11,15 @@
...
       <li><h4>Ensure a stable USB connection</h4>Put the watch and your computer on the same flat surface. Try to not move the cable connecting your computer and the watch during the whole flashing process. In case of worn or corroded contacts and pogo-pins that lost their springiness, it might be useful to attach the watch to its cradle or charging cable using duct tape or rubber bands.</li>
-      <li><h4>Repeat the flash process</h4>Some users have reported needing up to five flashing attempts for a successful result.<br>There is a small chance that the image files were corrupted during download. We do not currently provide checksums for the files. However, if you download the files again under a different filename and then compare them with <code>sha256sum filename</code>, the checksums must match.</li>
+      <li><h4>Repeat the flash process</h4>Some users have reported needing up to five flashing attempts for a successful result.<br>There is a small chance that the image files were corrupted during download. To guard against this, it's recommended to verify the checksums.  Download the  <code>SHA512SUMS</code> file (
+{{#if deviceName}}
```
FlorentRevest, line 15: Iiuc this condition is no longer needed now that https://github.com/AsteroidOS/asteroidos.org/pull/324 is merged ?

beroset:
Does something else need to be changed here?

PR #294, opened 2023-03-17
Install instruction selection via radio buttons

eLtMosen:
This PR 
- Implements a jQuery logic to selectively show install instruction based on user selection
- Moves the install instructions into an include block

Draft, todo:
- Currently only included into aw-install.hbs, to be implemented in all install pages
- mooneye -> temp install only
- Beluga -> warnings
- integration of above cases via handle bars as preparation for major consolidation of the templates which will be not part of this PR

Current status:
https://user-images.githubusercontent.com/15074193/225778083-c36241d3-d46c-4966-8383-456724c69d80.mp4


eLtMosen:
This has turned into a complete every line conflict with https://github.com/AsteroidOS/asteroidos.org/pull/324/files
Closing in favor of first consolidating the install scripts and then rebuilding this PR on top of that consolidation.

PR #301, opened 2023-03-19
Swap sturgeon for higher quality version

eLtMosen:
@MagneFire please check if you like this swap to non metal straps.

The pic with metal strap was not available in > 600px.


MagneFire:
I don't have a strong opinion on either. `sturgeon` was sold with either the black leather strap or the metal band.

Ps. Your commit message still mentions `anthias` : 😉

PR #307, opened 2023-03-19
Swap narwhal for higher quality version

eLtMosen:
Narwhal version with narrower straps and thus larger display.
Please @dodoradio  ,check if this is ok for you :D


dodoradio:
Looks great, thank you!

PR #323, opened 2023-03-24
Remove piwik analytics stuff from head

eLtMosen:
According to @jrtberlin on matrix "We don't have a service under that subdomain. We can remove all references to the analytics stuff"

### DIFF
--- a/templates/includes/head.hbs
+++ b/templates/includes/head.hbs
@@ -31,19 +31,3 @@
 
 <!-- Favicons -->
 <link rel="shortcut icon" href="{{ assets }}/ico/favicon.ico">
-
-
-<!-- Matomo -->
-<script type="text/javascript">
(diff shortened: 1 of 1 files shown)


FlorentRevest:
FWIW, we used to have Google Analytics and I'd sometimes like to peak at the stats to know how the community was growing. Then some users started to complain about the website sending requests to Google so we switched to a self-hosted Matomo instance. In theory, it'd have been nice, but like all our previous attempts at self-hosted service, it was just left to rot. I never looked into its statistics because the usability was terrible and we essentially stopped having analytics already. At some point, we migrated to a newer server and we didn't bother to migrate the Matomo endpoint because no one missed it anyway. Now we've been left with this dangling reference to a dead analytics server for a couple of years... :) 

Anyway, I guess we'll just not have statistics then ahah. :shrug:

PR #324, opened 2023-03-27
Consolidate templates

### LINKED ISSUE #292: Consolidate Templates by extending handle-bar use
This could be a good first step towards consolidating all these templates https://github.com/AsteroidOS/asteroidos.org/tree/master/templates/layouts which are out of control already.

Alternatively, I've been talking a lot about using conditionals like https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/includes/documentation-header.hbs#L5 to make templates more dynamic based on parameters (here the condition is on the slug variable: https://github.com/AsteroidOS/asteroidos.org/blob/master/pages/wiki/documentation.md?plain=1#L3 but we could use arbitrary variables. For example, we could have

```
{{#is codename 'beluga'}}
  {{> bootloader-unlock-beluga}}
{{else}}
  {{> bootloader-unlock-aw }}
{{/is}}
```

Or

```
{{#is watch-needs-simg 'yes'}}
        <a class="btn btn-primary" href="https://release.asteroidos.org/nightlies/{{deviceName}}/asteroid-image-{{deviceName}}.simg" role="button">asteroid-image-{{deviceName}}.simg</a>
{{/is}}
```

Or `{{#is two-columns 'yes'}}` (for firefish/ray) etc... There is a balance to be found between using different templates or include blocks. For example, maybe the mtk-install is so different from the aw-install that they shouldn't be folded together, but maybe they should, I'm not sure about that one. But ultimately the goal should just be to 1- avoid duplication 2- Keep things modular and easy.

For example, one thing I find quite bizarre looking back on our code (I should have noticed it at the time of the pull request review :) ) is our empty `watches.md` https://raw.githubusercontent.com/AsteroidOS/asteroidos.org/master/pages/main/watches.md I would expect the content of that page to be in `watches.md` because this is just an empty `body` that gets included in https://github.com/AsteroidOS/asteroidos.org/blob/master/templates/layouts/watches.hbs#L33 (weirdly, just after the content of the page is defined in the layout...)

Similarly, we could hide the navbar of the documentation landing page only when `{{#is slug 'documentation'}}`.

I've been restraining from being too annoying with this but I really think we're at the point where this sort of consolidation is needed to keep the website easy to evolve uh.

The language used for all this templating is called HandleBars https://handlebarsjs.com/guide/#what-is-handlebars

_Originally posted by @FlorentRevest in https://github.com/AsteroidOS/asteroidos.org/issues/285#issuecomment-1467110970_

beroset:
This consolidates all of the install templates and fixes #292 except for mtk-install into one aw-install template and separates some of the generic install instructions into includes.  There isn't technically any reason mtk-install couldn't also be folded in by using  the new `installParts` and `simg` variables but it didn't seem vital to do so at this moment.


FlorentRevest:
Fantastic! Thank you so much @beroset, I know that refactoring can sometimes feel unrewarding but that was very much needed and this unblock much easier improvement of our install pages :D

Issue #265, opened 2023-03-05
[Watches page] install-box divs do not reliably wrap into planned amount of colums


eLtMosen:
As first found by @MagneFire 

Attached screen width should result in 3 columns.
Reducing the install-box margins might mitigate that while pushing the list off center.
Another approch would be to use Flexbox via `display: flex`. Which is to be investigated.

![grafik](https://user-images.githubusercontent.com/15074193/222991215-668db73e-54a9-40d1-85bf-7455ffa4045b.png)

PR #346, opened 2023-04-10
watches: Reduce width on smaller screen widths.

MagneFire:
This ensures that there's enough space available such that there's no empty space on the right on the watches page.

Here's a before and after of the phenomenon.
image

### DIFF
--- a/styles/components/install-box.less
+++ b/styles/components/install-box.less
@@ -35,12 +35,12 @@
   }
 
   @media (min-width: @screen-sm-min) {
-    width: 31.2%;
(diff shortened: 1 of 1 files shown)


eLtMosen:
Very good! I was ment to fix this since i drove you into that misery :P
So thanks! Will test and it likely fixes https://github.com/AsteroidOS/asteroidos.org/issues/265

eLtMosen (approved):
Yepp, tested and works with minor changes to the visual appearance.

PR #350, opened 2023-04-13
Add stars column to featuretable

beroset:
In conversation with @eLtMosen, he suggested that it would be nice to have our stars ratings repeated in the features table.  This implements that suggestion.


eLtMosen (approved):
And it looks great as expected
grafik

PR #351, opened 2023-04-13
Fix featuretable alignment

### LINKED ISSUE #349: Feature table breaks at inharmony
grafik

beroset:
This fixes #349 by slightly enlarging the width of the name column to accomodate inharmony, which is currently the longest name.

### DIFF
--- a/styles/components/install-support-table.less
+++ b/styles/components/install-support-table.less
@@ -89,7 +89,7 @@ th {
 }
 
 .name-col {
-  min-width: 70px;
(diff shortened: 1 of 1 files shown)


eLtMosen (approved):
Yepp, valid fix until we get an even longer codename, ehehe
I tried % width but that looks awkward when the list gets wide. Which would make viewport sensitve valus necessary. Yaddi yadda, Lets just go with this quick fix.
Appreciate the swift action!

Issue #343, opened 2023-04-09
Slug is either too high or not centered to the mean line of the title


eLtMosen:
![image](https://user-images.githubusercontent.com/15074193/230767316-ec831903-d2a8-465e-9de8-5537070dd3fe.png)

eLtMosen:
I'd say it needs to be centered to the `()`
![image](https://user-images.githubusercontent.com/15074193/230803249-bf4f7334-9032-48ba-8e00-d526dc220281.png)

eLtMosen:
@MagneFire @FlorentRevest opinions on this?

- The slug is an svg and i would first of all remove the svg code into a svg file
- I would try to make look the slug consistent to the `()` in height, linewidth and vertical centering to the font mean line.

Practically, it will become a bit thicker and likely a tad taller to be as high as the `()`. 

Also, while i have you here, look at the (catfish/catfish_ext/catshark) cluster. It is visually quite appealing. But breaks the mobile view since its too wide and forces the viewport width to scroll horizontally.
https://github.com/AsteroidOS/asteroidos.org/issues/329

On that, i'd like to make the `, ` and ` / ` consistent with spacings properly used like in text grammar. And implement that over the whole website. This would impact the watches gallery page too for the comma separated multiple codename/variant watches.

MagneFire:
> @MagneFire @FlorentRevest opinions on this?
> 
>     * The slug is an svg and i would first of all remove the svg code into a svg file
> 
>     * I would try to make look the slug consistent to the `()` in height, linewidth and vertical centering to the font mean line.
> 
> 
> Practically, it will become a bit thicker and likely a tad taller to be as high as the `()`.

I'm fine with it being a bit thicker. Making it consistent with the brackets seems the more important part imho.

> 
> Also, while i have you here, look at the (catfish/catfish_ext/catshark) cluster. It is visually quite appealing. But breaks the mobile view since its too wide and forces the viewport width to scroll horizontally. #329
> 
> On that, i'd like to make the `, ` and `/` consistent with spacings properly used like in text grammar. And implement that over the whole website. This would impact the watches gallery page too for the comma separated multiple codename/variant watches.

We could also remove the codename from the header since there's already a table below it containing the watch codenames. Else the idea of using consistent spacing a good idea as well.
Having the watches page comma separated looks just as good to me:
[...]

PR #353, opened 2023-04-14
Add unicode slug to watches, news and documentation header

eLtMosen:
Fixes https://github.com/AsteroidOS/asteroidos.org/issues/343 by swapping the svg slug for unicode char U+27E9.

grafik
grafik
grafik

### DIFF
--- a/templates/includes/documentation-header.hbs
+++ b/templates/includes/documentation-header.hbs
@@ -4,11 +4,7 @@
     <h1>
     {{#is slug 'documentation'}}
     {{else}}
-      <a href="{{rel 'wiki/documentation'}}">Documentation</a> <svg width="4.3482mm" height="6.0193mm" version="1.1" viewBox="0 0 6.9570877 9.6308298">
-<g transform="translate(-73.598 -103.17)">
-<g transform="translate(1.6)" fill="#666">
-<path d="m73.598 103.17h0.85725l2.8998 4.8154-2.8998 4.8154h-0.85725l2.8892-4.8154z" fill="#666" stroke-width=".26458"/>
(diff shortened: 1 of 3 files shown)


templates/includes/documentation-header.hbs
```diff
@@ -4,11 +4,7 @@
     <h1>
     {{#is slug 'documentation'}}
     {{else}}
-      <a href="{{rel 'wiki/documentation'}}">Documentation</a> <svg width="4.3482mm" height="6.0193mm" version="1.1" viewBox="0 0 6.9570877 9.6308298">
-<g transform="translate(-73.598 -103.17)">
-<g transform="translate(1.6)" fill="#666">
-<path d="m73.598 103.17h0.85725l2.8998 4.8154-2.8998 4.8154h-0.85725l2.8892-4.8154z" fill="#666" stroke-width=".26458"/>
-</g></g></svg> {{/is}}
+      <a href="{{rel 'wiki/documentation'}}">Documentation</a> ⟩ {{/is}}
```
beroset, line 7: I like it, but that unicode symbol might be hard to spot for people doing maintenance on these handlebars files.  Might it be better to use `&#x27e9;` instead?

beroset: On the other hand, sources I've found recommend using the character directly if the encoding is UTF-8.  It doesn't look like we explicitly set the encoding, however, so I'm not sure.

eLtMosen: Agreed, bad discoverability while maintaining is a striking argument.

beroset (approved):

Issue #240, opened 2023-02-23
Fix clipping of the sidebar when headers are long and result in multiple lines in the sidebar


eLtMosen:
(although this clipping is a bug and should be fixed separately)

_Originally posted by @FlorentRevest in https://github.com/AsteroidOS/asteroidos.org/issues/236#issuecomment-1442537288_
            
            
![image](https://user-images.githubusercontent.com/15074193/221050344-67bb4053-8254-4b8e-a2ef-f67cfa9198f5.png)

eLtMosen:
Keeping this open despite the mitigating merge of https://github.com/AsteroidOS/asteroidos.org/pull/355#pullrequestreview-1386841009
Since the issue is still present in the Documentation -> Building Asteroid pages for instance.
And someone actually fixing the bootstrap sidebar behavior with long lists would be very welcome.

eLtMosen:
was fixed in #355 but forgotten to close here

PR #355, opened 2023-04-16
[FAQ] Sort entries into sub categories

eLtMosen:
Trying to solve this issue https://github.com/AsteroidOS/asteroidos.org/issues/240 now since weeks, i came to the realization that it would involve to mess with bootstrap. Since the sidebar handling and thus buggy addition of the `top:` parameter that should prevent the sidebar from overlapping with the footer is a bootstrap thing.

This PR aims to sort the entries into sub headings to work around this issue. By using the sidebar like intended, with sub categories shortening the sidebar height and thus prevent it to break at the footer.

We can sense-fully sort the entries like this imho since the resulting categories would then thematically fit the sub heading entries.
- What is AsteroidOS 
- Is Asteroid OS based on Android
- Can my watch run AsteroidOS
- Question not answered

This is how it would behave:

https://user-images.githubusercontent.com/15074193/232300326-d1c48e21-eb91-4915-b27b-91e721f0ab03.mp4


beroset (approved):
I think this is a good change.

MagneFire (approved):
Good change!

PR #359, opened 2023-04-27
Translate hint

beroset:
This fixes https://github.com/AsteroidOS/asteroid-settings/issues/80 by adding a language translation hint to the tips and tricks section of the "creating an AsteroidOS app" page.  It also enhances the command line build script in that same section.


eLtMosen (approved):
That was quick!
And another issue closed thanks to @beroset

PR #363, opened 2023-05-02
Clarify boot troubleshooting advice

beroset:
Based on my own recent confusion, this clarifies the instructions for how to troubleshoot boot problems by providing exact commands and clarifying the context in which the commands are issued.


eLtMosen (approved):
Classic case of "Too many trees to see the wood" :D
Tested just now to double confirm.

Issue #366, opened 2023-05-08
Provide more detailed SoC information.


MagneFire:
https://github.com/AsteroidOS/asteroidos.org/commit/ec6737fd00276fffbecb217cba2cf503d13fbfa0 added SoC information for all supported watches.
Currently this is very limited for some watches. There are different `msm` based watches that can either use an `msm8226`, `msm8909w`, `msm8937`, etc.

We should figure out what different SoCs are used and update the information in our 'database' accordingly.

PR #367, opened 2023-05-08
Add more detailed info on MSM SOC

beroset:
This adds more detail to the SOC versions for MSM variants, specifically naming either msm8909 or msm8226 based on the CONFIG_ARCH_MSM8xxx=y line in each watch kernel's defconfig.

This fixes https://github.com/AsteroidOS/asteroidos.org/issues/366

### DIFF
--- a/pages/watches-support.json
+++ b/pages/watches-support.json
@@ -12,7 +12,7 @@
       "maintainers": [ ],
       "contributors": [ "locusf", "MagneFire", "kido" ],
       "stars":3,
-      "soc":"msm",
+      "soc":"msm8226",
       "kernelversion":"3.10",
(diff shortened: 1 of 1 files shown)


pages/watches-support.json
```diff
@@ -71,7 +71,7 @@
       "maintainers": [ "MagneFire", "wannaphong" ],
       "contributors": [ "MagneFire", "wannaphong" ],
       "stars":3,
-      "soc":"msm",
+      "soc":"msm8909",
```
MagneFire, line 74: Technically all Wear 2100 and  Wear 3100 watches use a `msm8909w`.

dodoradio:
looks great, thank you for doing this so quickly!

PR #372, opened 2023-05-12
add rss feed to footer, generated by politepol

rmirau:
i still like to use rss to stay updated on news and projects that are of interest, therefore i created an rss feed for the news blog of asteroidos.org.
i've used the service the first time now, so i'll have to make sure that the feed is updated automically when a new blogpost is released later on. i'm optimistic that further changes won't be necessary though


FlorentRevest:
Thank you @rmirau for the contribution! I agree it could be nice to have an RSS feed here but, if possible, I'd rather not depend on a third party website for that.

It feels that we could rather easily generate an RSS feed with a small variation of the news page https://github.com/AsteroidOS/asteroidos.org/blob/master/pages/main/news.md?plain=1#L6 ? Instead of generating an HTML page, we could use that same Handlebars loop to generate an RSS XML feed. Then we could host our own RSS feed.

Maybe other maintainers have other opinions.

beroset:
I agree that an news feed could be done that way, and it's probably better not to rely on a third party that may or may not exist in the future.  I'd be happy to code this up but what version to use?

1. the [RSS Advisory Board spec](https://www.rssboard.org/rss-specification) from 2009
2. the [IETF Atom spec](https://datatracker.ietf.org/doc/html/rfc4287) from 2005
3. the [W3C HTML5 `article` element](https://html.spec.whatwg.org/multipage/sections.html#the-article-element) from 2023

eLtMosen:
The `article` element looks quite neat tbh. Thanks for linking that up in any case.

MagneFire:
https://github.com/AsteroidOS/asteroidos.org/pull/409 has been merged. The RSS feed is now available at: https://asteroidos.org/feed.xml

PR #380, opened 2023-05-23
Add watch technical details page

beroset:
This adds a technical details page that lists the SOC, kernel version and Android version for each watch, both experimental and supported.


pages/wiki/technical.md
```diff
@@ -0,0 +1,65 @@
...
+
+<table class="table table-striped">
+  <tr>
+  <thead>
+    <th>Model Name</th>
+    <th>Codename</th>
+    <th>Support</th>
+    <th>SOC</th>
+    <th>Kernel</th>
+    <th>original Android</th>
```
eLtMosen, line 24: Original Android, with uppercase start imo.
And i am not totally sure if it fits perfectly, but "Stock Android" might be shorter .
Afaiu, the term Stock Android describes the android version supplied by the manufacturer or google and is often used to refer to the baseline or reference version of the operating system.

pages/wiki/technical.md
```diff
@@ -0,0 +1,65 @@
+---
+title: Technical Details
+layout: documentation
+---
+This page contains technical details about currently supported watches.
+
+# AsteroidOS Watch Technical Details
+---
+
+While AsteroidOS does not use Android, it does rely on drivers that are be supported with [libhybris](https://github.com/libhybris/libhybris).
```
eLtMosen: "that are be supported" to
"that are supported when using" (Which would hint at mainline not using libhybis) or
"that are supported by"

Or a completely different approach, imo clearer:
"Although AsteroidOS does not use Android, it relies on libhybris to use the Android drivers."

pages/wiki/technical.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Supported watches
+
+<table class="table table-striped">
+  <tr>
+  <thead>
+    <th>Model Name</th>
+    <th>Codename</th>
+    <th>Support</th>
+    <th>SOC</th>
```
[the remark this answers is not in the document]
beroset: Good point.  I addressed that and also added a note explaining what the term means.

eLtMosen (commented):
Great initiative.
Some things i found while reading trice.Requesting change only for the grammar hickup.

pages/wiki/technical.md
beroset: Good point.  I changed it to "While AsteroidOS does not use Android, it does rely on Android drivers that are, in the context of AsteroidOS, supported with [libhybris](https://github.com/libhybris/libhybris)."

pages/wiki/technical.md
```diff
@@ -0,0 +1,65 @@
+---
+title: Technical Details
+layout: documentation
+---
+This page contains technical details about currently supported watches.
+
+# AsteroidOS Watch Technical Details
```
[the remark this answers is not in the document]
beroset: I went with "Technical Details of AsteroidOS Watches" and removed line 5.  Thanks!

pages/wiki/technical.md
beroset: I put "Original Android" and also added a link to the part of the FAQ that describes the relationship to Android drivers in more detail.

eLtMosen:
Thanks for the swift action to everyone involved!

z3ntu:
It's beautiful! :heart_eyes: https://asteroidos.org/wiki/technical/

PR #381, opened 2023-05-23
Correct soc value for wren

z3ntu:
Wren, just like its sibling sparrow, is a msm8226 device.


MagneFire:
Good catch! Thanks for the PR.

PR #384, opened 2023-06-02
[Features table] Update pike and bump last-updated date

eLtMosen:
- Change GPS to good on pike
- Bump last updated section date
- Add spacing to footer

### DIFF
--- a/pages/watches/features.md
+++ b/pages/watches/features.md
@@ -3,8 +3,6 @@ title: Feature list
 slug: features
 layout: featuretable
 ---
-<br>
-<h4>This table has been updated on: <code>2023-05-13</code></h4>
 <h2>Legend</h2>
 <table>
 <tr><td class="name-col"><b>good</b></td><td class="legend-col good" /><td>The hardware exists and is working</td></tr>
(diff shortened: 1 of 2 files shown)


pages/watches/features.md
```diff
@@ -13,6 +11,11 @@ layout: featuretable
...
-
-<br/>
-For more information on each device, click on the respective code name in the first column. These are linked to the entries on the <a href="{{rel 'watches'}}">Watches</a> page.
+<br>For more information on each device, click on the respective code name in the first column. These are linked to the entries on the <a href="{{rel 'watches'}}">Watches</a> page.
+<br>
+<h3>This table has been updated on:</h3>
+<code>2023-06-02</code> (Change GPS on pike to good)
+<br><code>2023-05-22</code> (Add koi/ayu)
+<br><code>2023-05-22</code> (Change WLAN on narwhal to good and Hands to partial)
+<br><code>2023-04-23</code> (Add hoki)
```
beroset, line 20: I have two questions about this.  First, does a reader really care?  It might be useful for maintainers, but we have git logs for this kind of thing.  I think I prefer just the "last updated" date.

Second, if we do decide to keep a table, wouldn't it be better as... a `<table>`?

beroset (approved):
Nothing major here, but I do have some questions.

pages/watches/features.md
dodoradio: I'm also not sure this is needed. I think the date gives enough of an idea of 'things are happening' and the detailed changes aren't really needed.
This is probably trying to reduce the number of 'is the project actually alive' questions, which is a noble cause, but I'm not sure this will actually solve it. I feel those people will always show up unless we're a household name in the linux distro world :sweat_smile:  - I think a date is sufficient here for this purpose

eLtMosen: Agreed and reverted back to only show the last updated date.

PR #385, opened 2023-06-03
Koi: mark as experimental due to bricking issues

dodoradio:
This marks koi as experimental and adds a warning about bricking issues 

my ayu just bricked itself, and I believe it is prudent to mark experimental until we figure out how fatal this is.
This is really really really disappointing.


dodoradio:
image
the 'callouts' section for this watch is now 3 items long and looks a bit absurd

beroset (approved):
Based on your recent experiences, this seems like a good amendment.

PR #386, opened 2023-06-06
[Watches page] Add spacing to footer

eLtMosen:
- Add hardcoded line breaks as spacing to footer. Using a template has been investigated but no universal solution was found.


FlorentRevest:
Thanks for the investigation ;)

CSS is hard. :p

PR #390, opened 2023-06-11
[Building asteroidos] Move source hint from codeblock

eLtMosen:
- The hint to not only run the prepare-build.sh but source it made the codeblock look confusing
- Double-click selection of the whole line contained the comment/hint.
- Moved over the codeblock
- Added important note to highlight it.


eLtMosen (commented):
Seems no obligations, merging ;)

PR #394, opened 2023-06-26
[Copy to clipboard button] Make function more universal 

eLtMosen:
- Remove the necessity to target IDs but find the closest install-code-wrapper class, find the code block in that and copy its content to clipboard.
- Remove redundant IDs from install pages code blocks since the copy to clipboard function does not rely on those anymore.

Signed-off-by: Timo Könnecke koennecke@mosushi.de"


MagneFire (approved):
This is amazing! Thanks for this improvement!
Tested, works well.

PR #395, opened 2023-07-04
watches: add speaker entry for hoki

r3pwn:
hoki has a speaker, but it looks like the feature entry for it may have been missed


MagneFire:
Thanks for this PR!

Can you change the status to `bad`? It won't work until we get the modem to boot. (https://github.com/AsteroidOS/meta-smartwatch/issues/185)

r3pwn:
@MagneFire Done :)

MagneFire:
Thanks!

PR #398, opened 2023-08-04
Add link to the wiki as a post install read.

MagneFire:
To increase exposure of the wiki add a link to it as a post installation step.


beroset (approved):
Good change!  Thanks.

PR #399, opened 2023-08-04
Add link to watch specific page.

MagneFire:
The idea being that one can inform themselves prior to installing AsteroidOS.

The only page that doesn't exist at the time of writing is the wiki page to `harmony/inharmony`.

| | |
|-|-|
| image|image|


eLtMosen (approved):
Good addition!

dodoradio (approved):

PR #400, opened 2023-08-14
Improve android-tools installation instructions

nerdius:
These changes add/fix `android-tools` installation instructions for openSUSE & SUSE Linux Enterprise, and improves overall document structure for various Linux distributions.

### DIFF
--- a/styles/components/install-preparation-box.less
+++ b/styles/components/install-preparation-box.less
@@ -13,7 +13,7 @@
 
   border: none;
   border-radius: 4px;
-  min-height: 485px;
+  min-height: 510px;
   display: inline-block;
   width: 100%;
 
--- a/templates/includes/install-prepare-adb.hbs
+++ b/templates/includes/install-prepare-adb.hbs
@@ -2,7 +2,7 @@
         <h3>Install ADB & Fastboot</h3>
         <br>
         <div name="LinAdb" style="display: block">
-          <h4>Debian, Ubuntu, etc.</h4>
+          <h4>Debian / Ubuntu / Mint</h4>
           <div class="install-code-wrapper" style="justify-content: center;">
             <div class="install-code-pre-wrapper">
               <pre class="install-code-pre" style="border-radius: 4px;"><code>sudo apt install android-tools-adb <br>android-tools-fastboot</code></pre>
@@ -11,7 +11,7 @@
               <input type="button" class="btn btn-primary clipboard-button" value="&#10697;"></input>
             </div>
           </div>
-          <h4>Fedora, CentOS, RHEL, etc.</h4>
+          <h4>Fedora / CentOS / RHEL</h4>
           <div class="install-code-wrapper" style="justify-content: center;">
             <div class="install-code-pre-wrapper">
               <pre class="install-code-pre"><code>sudo dnf install android-tools</code></pre>
@@ -20,7 +20,7 @@
              <input type="button" class="btn btn-primary clipboard-button" value="&#10697;"></input>
             </div>
           </div>
-          <h4>openSUSE, SLE</h4>
+          <h4>openSUSE / SLE</h4>
           <div class="install-code-wrapper" style="justify-content: center;">
             <div class="install-code-pre-wrapper">
               <pre class="install-code-pre"><code>sudo zypper install android-tools</code></pre>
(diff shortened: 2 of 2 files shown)


templates/includes/install-prepare-adb.hbs
```diff
@@ -2,7 +2,7 @@
         <h3>Install ADB & Fastboot</h3>
         <br>
         <div name="LinAdb" style="display: block">
-          <h4>Debian / Ubuntu</h4>
+          <h4>Debian, Ubuntu, etc.</h4>
```
eLtMosen, line 5: Visually, i am not much a fan of unnecessarily repetitive elements like the `etc.` in multiple lines.
For this line, i'd suggest to keep the  `Debian / Ubuntu` since everyone using a derivative distro is very likely familiar with those names and the `etc.`  does not add much here . Also, this is supposed to be a headline and it looks unfortunate to end a headline on a `.` imo.

templates/includes/install-prepare-adb.hbs
```diff
@@ -11,7 +11,7 @@
               <input type="button" class="btn btn-primary clipboard-button" value="&#10697;"></input>
             </div>
           </div>
-          <h4>Fedora / OpenSUSE / CentOS</h4>
+          <h4>Fedora, CentOS, RHEL, etc.</h4>
```
eLtMosen, line 14: Agreed about adding RHEL.
For heading context, i'd like to avoid `,` and `.`
The `/` better indicates a difference to body text imo.

templates/includes/install-prepare-adb.hbs
```diff
@@ -20,7 +20,16 @@
...
+          <h4>openSUSE, SLE</h4>
+          <div class="install-code-wrapper" style="justify-content: center;">
+            <div class="install-code-pre-wrapper">
+              <pre class="install-code-pre"><code>sudo zypper install android-tools</code></pre>
+            </div>
+            <div class="clipboard-button-wrapper">
+             <input type="button" class="btn btn-primary clipboard-button" value="&#10697;"></input>
+            </div>
+          </div>
+          <h4>Arch, Manjaro, EndeavourOS, etc.</h4>
```
eLtMosen, line 32: When writing `Arch and derivatives`, i was assuming every Arch user knows. But indeed might be helpful and looks informative to add Manjaro and EndeavourOS.
Tho the EndeavourOS docs advise to use `yay` for newbies needing the hint that they are arch based, hehe.

eLtMosen (changes requested):
Thank you very much for this swift PR. And sorry about not noticing earlier and doing another PR here. https://github.com/AsteroidOS/asteroidos.org/pull/401
I have reverted mine to a draft and kept it just to illustrate the height problem we introduce by adding another paragraph into the box.

The change-request is about reverting the `,` back to `/` and omitting the `etc.`.

Regarding the install-preperation-box height, those are supposed to be equal height for the download related boxes since they are placed next to each other.

grafik

Adapting the height is just a suggestion, not a blocker for this PR.
I can rework that in a follow up PR.

nerdius:
@eLtMosen, I've added two more commits here.

I see the issue with the height. Using `min-width` could be good enough in this case, but there's a better general approach with flexbox. Although, it requires some extra markup for each row, so maybe it's not worth the effort.

I've changed commas bach to slashes. I personally think commas are better stylistically, but you should know better. (It's probably about _AND_ vs. _OR_.)

Regarding the categorization, I used _etc._ just because it's a list and there's always more in Linux! I see the issue with repetition. Using _and their derivatives_ also has the same issue, and it's _way_ longer. Another approach is to just use the distribution family, like this:

- Debian family
- RHEL family
- SUSE family
- Arch family

But this also has a repetition, and it's too generic, technical, and not newbie-friendly.

So I decided to list 2-3 most popular distros in each category.

Regarding `yay` vs. `pacman` for Arch derivatives, while that's also my favourite way, it might not be available by default on some more niche Arch derivatives as they might use something else or nothing at all (like Arch by default). Besides, that level of detail is unnecessary.

styles/components/install-preparation-box.less
```diff
@@ -13,7 +13,7 @@
 
   border: none;
   border-radius: 4px;
-  min-height: 485px;
+  min-height: 510px;
```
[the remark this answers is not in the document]
nerdius: No problem. Yes, I'll look into this.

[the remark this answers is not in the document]
FlorentRevest: I'll proceed to merge eLtMosen's PR since it's ready but thanks for your contribution!

PR #402, opened 2023-08-17
[unlock] Warn to not relock

eLtMosen:
- Make clear in the info box to not relock the bootloader while AsteroidOS is installed
- Extend first sentence by <and booting> to make clear that AsteroidOS always needs an unlocked BL

grafik


MagneFire:
LGTM, but can you change "Bootloader" to "bootloader"?

PR #407, opened 2023-09-16
beluga: fix codenames

argosphil:
* fixes https://github.com/AsteroidOS/meta-smartwatch/issues/225


FlorentRevest:
My understanding is that this is not yet 100% sure right ? Change looks fairly benign but let's merge this when we know for sure :)

PR #408, opened 2023-09-21
Generate RSS feed via internal loop and add the link to footer.

lepras:
I am not that sure about how the `grunt` works, but I tried my best to create the hbs' that will get us a nice rss at `/rss`.

I wanted a rss for my feed clients, so referred [this](https://github.com/AsteroidOS/asteroidos.org/pull/372) pull request. Rather than asking about uPdAte, I decided to make a PR.

I tried testing it by building it locally according to [this](https://wiki.asteroidos.org/index.php/Local_version_of_asteroidos.org#Running_locally) guide. It didn't got me any errors.


MagneFire:
I've tried this out and expanded on some issues: https://github.com/MagneFire/asteroidos.org/tree/rss

But that work results in the following issues:
- Plugins need to be disabled as the HTML sanitizer adds `<html><head></head><body></body></html>` tags.
- The content type is still set to HTML potentially causing issues with clients.

https://github.com/AsteroidOS/asteroidos.org/pull/409 solves these by generating a dedicated `feed.xml` which isn't sensitive to the mentioned issues.

MagneFire:
https://github.com/AsteroidOS/asteroidos.org/pull/409 has been merged. The RSS feed is now available at: https://asteroidos.org/feed.xml

PR #409, opened 2023-09-27
Generate RSS feed from news items.

MagneFire:
This generates an RSS page for use with RSS readers.

I've tested this on various RSS reader with mixed results:
- Feeder (Android): Works, including images
- Newsboat: Works
- NewsFlash: ~~Feed remains empty~~ Works fine when not using a remote URL.

~~Currently marking as draft for the above reason and the fact that I had to downgrade some packages as the imported code is old and using the current packages didn't seem to work initially.~~ Issues have been resolved.

This PR supersedes: https://github.com/AsteroidOS/asteroidos.org/pull/372 and https://github.com/AsteroidOS/asteroidos.org/pull/408.


lepras:
merge wen? 🥹

MagneFire:
> merge wen? 🥹

I'm currently waiting for feedback on the PR before merging it.

It can take some time for the other maintainers to pick this up ;)

FlorentRevest:
Thanks a lot! This was more involving than I originally thought but at least now it's clean and we don't need to think about it anymore! :)

MagneFire:
> > we don't need to think about it anymore!
> 
> How are the commits pushed to production? I don't see any github action.

It was an issue on or website server. It has been fixed and the feed is now available: https://asteroidos.org/feed.xml

FlorentRevest:
> > we don't need to think about it anymore! 
> 
> How are the commits pushed to production? I don't see any github action.

We don't use GitHub actions, we have a bot that subscribes to GitHub notifications and one of the things it does with it is to rebuild and deploy the server.

Issue #410, opened 2023-11-02
Add release date of original watch in feature list


Mr-Bajs:
I suggest that the release year of the first model of a watch is added in the feature list. That would in addition to new potensial user to sort what kind of watch he or him might want, and make a better choice in terms of price, availability and level of new tech. Some users maybe dont want older tech, while another does care and might find an older watch more intressting due to its probably much cheaper.

I think that data point would have some value.

Saarsk:
I've been missing this every time I look through the watch list. And then eventually sort of stopped browsing because searching up specs for each model individually (all while remembering all of it for each model) is just too tedius and time-consuming.

And on the next occations I revisited the AsteroidOS site to see if there had been any progress, I'd forgotten a lot of the spec info I read about to compare the watches.

And so I never really got around to getting a watch to put AsteroidOS on. Partly also because I feel like the watches are old and also limited in terms of feature and hardware support parity with WearOS.

beroset:
The release date is now part of the wiki:  https://wiki.asteroidos.org/index.php/Technical_Details_of_AsteroidOS_Watches

PR #411, opened 2023-11-25
Add the release date of the watch

Tydragon00:
I agree with this [issue](https://github.com/AsteroidOS/asteroidos.org/issues/410) but I think it would be convenient to know which ones are the most recent without opening the individual watch page.
For some watches I did not add the date because they do not correspond to a single watch

### DIFF
--- a/pages/watches-support.json
+++ b/pages/watches-support.json
@@ -16,6 +16,7 @@
       "soc":"msm8226",
       "kernelversion":"3.10",
       "androidversion":"marshmallow",
+      "releasedate": 2014,
       "features": [
         { "name":"Display", "status":"good" },
         { "name":"Touch", "status":"good" },
@@ -45,6 +46,7 @@
       "soc":"msm8226",
       "kernelversion":"3.10",
       "androidversion":"lollipop",
+      "releasedate": 2015,
       "features": [
         { "name":"Display", "status":"good" },
         { "name":"Touch", "status":"good" },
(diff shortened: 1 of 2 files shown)


pages/watches-support.json
```diff
@@ -634,6 +650,7 @@
       "soc":"msm8226",
       "kernelversion":"3.10",
       "androidversion":"marshmallow",
+      "releasedate": 2015,
```
FlorentRevest, line 653: Maybe it's worth removing the "2015" from the model name here ("Moto 360 2015") ?

templates/includes/installbox.hbs
```diff
@@ -1,7 +1,10 @@
...
-  <a href="{{#if reference}}{{reference}}{{else}}{{name}}{{/if}}"><div class="install-box">
+<a href="{{#if reference}}{{reference}}{{else}}{{name}}{{/if}}">
+  <div class="install-box">
     <img src="../public/img/{{name}}.png" class="install-watch-img"><br>
     <b>{{models}}</b><br>({{name}}{{#if othernames}}{{#othernames}}, {{.}}{{/othernames}}{{/if}})<br>
-    <i>Support: {{#starString stars 5}}{{/starString}}</i>
-  </div></a>
-{{/with}}
+    <i>Support: {{#starString stars 5}}{{/starString}}</i><br>
+    {{#if releasedate}} <i>Release date: <b>{{releasedate}}</b></i>{{/if}}
```
FlorentRevest, line 7: I fear this would take quite a bit of space in the install box, which we've been trying to keep concise. How about a small hint in parenthesis after the name ? It should be obvious that the year is the release date.

@eLtMosen any opinions on this ?

beroset:
We now have the release date as part of the wiki:  https://wiki.asteroidos.org/index.php/Technical_Details_of_AsteroidOS_Watches

PR #425, opened 2024-06-17
Updated watches-support.json for Hoki features

855579762735:
updated feature statuses for the Hoki entry to include good support for GPS, NFC, Tilt-to-Wake, and Heart Rate.


eLtMosen:
Thanks for testing and updating here!

PR #427, opened 2024-07-11
[catfish] Add info notes for GPS and LCD support

eLtMosen:
As noted by hacker420 in the matrix chat, an explanation to why GPS and LCD are listed as partially supported might be good for clarity.

After the change:
grafik

### DIFF
--- a/pages/watches/catfish.md
+++ b/pages/watches/catfish.md
@@ -7,6 +7,6 @@ layout: aw-install
 ---
 <div class="callout callout-info">
     <p>Hardware support notes:</p>
-    <p>Dual-Display support<br> Is marked partial due to incomplete support. The secondary LCD works on all models and shows the time and steps when enabled. However, enabling the step counter and the timepiece mode is currently not available through the user interface.</p>
+    <p>Dual-Display support<br> Is marked partial due to incomplete support. The secondary LCD works on all models and shows the time and steps when enabled from cli. However, enabling the step counter and the timepiece mode is currently not available through the user interface.</p>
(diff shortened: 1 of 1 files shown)


pages/watches/catfish.md
```diff
@@ -5,3 +5,8 @@ label: catfish/catfish_ext/catshark
 section: install
 layout: aw-install
 ---
+<div class="callout callout-info">
+    <p>Hardware support notes:</p>
+    <p>Dual-Display support<br> Is marked partial due to incomplete support. The secondary LCD works on all models and shows the time and steps when enabled. However, enabling the step counter and the timepiece mode is currently not available through the user interface.</p>
```
dodoradio, line 10: I'd add '... shows time and steps when enabled from cli.' just to hint at how it's possible to enable it for users who aren't aware.
otherwise, looking good!

dodoradio (changes requested):

pages/watches/catfish.md
Hacker1245: possibly add with what commands and switches? Something like: "You can enable step counting with 'lcd-tools --enable-stepcounter'"

eLtMosen: > possibly add with what commands and switches? Something like: "You can enable step counting with `lcd-tools --enable-stepcounter`"

I think that is out of scope of the install pages and belongs to the affected watches wiki entry?

MagneFire: I think a adding such information to a wiki page is indeed better.

MagneFire (approved):

PR #433, opened 2024-08-21
Remove documentation and faq

MagneFire:
The documentation and the FAQ is now available via a MediaWiki instance at wiki.asteroidos.org. Redirects are enabled for the removed wiki and FAQ pages.

Leaving this PR open to invite discussion on keeping the documentation for archiving purpose (the git history already does this imo).


beroset:
I think it's a good idea to remove it.  If we would ever need the old state of these for some reason, I agree, yes, that's what the git repository is for.

PR #436, opened 2024-12-12
Use css animation, not obsolete -webkit-animation

beroset:
This updates the slide show animation (for the screen images at the top of the main page) to use animation which is now supported by all browsers, rather than the deprecated -webkit-animation.

It also changes the animation so that it goes forward and then reverses rather than its current behavior of resetting quickly to the beginning.


eLtMosen:
Hehe. stuff deprecates faster than we act. Thanks a lot for the welcome maintenance! I like the change to scroll backwards a lot.

Issue #423, opened 2024-04-28
Revise watch rating system 


dodoradio:
Currently the rating system does not really reflect what watches we recommend to new users. We frequently recommend catfish and beluga, while the top two watches on the ratings page are still bass and sturgeon. I think it would be wise to revise the rating system (from what was introduced in https://github.com/AsteroidOS/asteroidos.org/issues/80 ) to something that reflects our current favourite.

dodoradio:
My proposal:
First star is awarded for a watch that is in 'proof of concept' state:
- boots
- must have working gui and touch, launcher is navigable

Watches cannot score higher than two stars unless they have:
- bluetooth
- fully working suspend, including adequate battery life
- if disassembly/dock required, assembly process/dock is not adequately documented and 3d printing files are not made available

Watches cannot score higher than three stars if:
- flashing requires disassembly, building a custom dock or other significant work
- serious hardware issues, such as on koi

Third star is for if a watch satisfies some, but not all of the conditions needed for fourth star

Fourth star:
- working WiFi
- most sensors present on the watch which are suppported by asteroidos are working (device-specific weird sensors and peripherals are not required)
- always on display, or equivalent, works
- minor UI stutters are allowed

Fifth star:
- Mostly subjective
- influenced by good battery life, UI fluidity, presence of GPS
- AsteroidOS Recommends!

dodoradio:
I'd also add the categories for these ratings to the wiki so that users have a better understanding of what we assess

eLtMosen:
Having skimmed through all watches and calculating the star rating according to the new proposed rating rules, i have no obligation against these changes and welcome the added clarity and traceability.

Many users showed slight confusion over the current rating system, especially about the fact that rather old watches with few hardware features have a better rating than newer and available ones.
Excluding watches that lack GPS from 5 stars is a good way to highlight catfish and beluga as our currently recommended go to acquisitions. 
Making bluetooth clearly a requirement for the 3 stars daily driver level is only pushing pike back to 2 stars for now. But clearly makes a cut for future ports to be able to sync with a phone to be a serious daily driver. Which i welcome for clarity.

FlorentRevest:
Makes sense to me too yes :) Thanks!

eLtMosen:
We have a support level page in the wiki now:
https://wiki.asteroidos.org/index.php/Watch_Support_Level

PR #441, opened 2025-02-01
Update watches-support.json to reflect new rating system

dodoradio:


eLtMosen:
Approval from my side. Cheers :D

MagneFire:
For future reference this is revised rating was suggested in https://github.com/AsteroidOS/asteroidos.org/issues/423

pages/watches-support.json
```diff
@@ -561,7 +561,7 @@
       "status":"supported",
       "maintainers": [  "jrt", "MagneFire" ],
       "contributors": [ "flocke", "fosspill", "FlorentBrianFoxcorner", "jrt", "MagneFire"  ],
-      "stars":3,
+      "stars":4,
```
MagneFire, line 564: Performance is really bad, it should not have an equal rating to for example `sturgeon`.

dodoradio: fixed

PR #443, opened 2025-02-06
Add ability to finely tune watch ratings and tune watches to better represent our recommendations

dodoradio:
following the recent change f99348e6238156df5aa7f86b9c948b327a8e0d52 we've finally re-rated our watches. However, catfish is still in second place instead of first. this PR adds the ability to use floats for the star values to more finely tune the positions of various watches. 
the second commit tunes the ratings of some watches which I picked based on personal opinion - this may need further review
Beluga is derated for worse repairability compared to catfish
sturgeon is rated higher due to excellent hardware
dory is derated due to low-res display and lack of wifi
skipjack is rated higher due to almost-4star support (tunny has a touchscreen bug which drags down skipjack)


dodoradio:
image

PR #444, opened 2025-02-06
Fix 'loose' instead of 'lose' typos in bootloader unlock instructions

dodoradio:


eLtMosen (approved):
Yay, thanks for fixing those!

PR #448, opened 2025-03-05
Add more prominent warning about making a backup

### LINKED ISSUE #447: Backup warning needs to be much more prominent
Although there are instructions guiding users to back up before doing a permanent install, it seems to be frequently overlooked and the consequences are not spelled out.

beroset:
Several people recently have asked on Matrix for help restoring their original OS but had failed to make a backup as mentioned on the installation instructions.  This adds a much more prominent warning and mentions the consequence of not backing up.  This fixes #447.


eLtMosen (approved):
That should be obvious enough now. lgtm

PR #474, opened 2026-02-08
rubyfish: Add watch to the supported list

MagneFire:
Currently draft PR until we have support merged and builds are available.


pages/watches-support.json
```diff
@@ -610,6 +610,44 @@
...
       ]
     },
+    {
+      "name":"rubyfish",
+      "wiki": "https://wiki.asteroidos.org/index.php/Rubyfish",
+      "models": [
+        "TicWatch Pro 3"
+      ],
+      "modelnumbers": [
+        { "num":"WH12018", "name":"TicWatch Pro 3 GPS", "codename":"rubyfish" },
```
MagneFire: This stuff probably needs another check. I've checked `WH11013` to match my watch, which should be the non-pro version with LTE.

moWerk (commented):
Lgtm.
The watch image is not scaled to proportional size of the other watches in the gallery yet. It shpould be larger and thus the straps squished to make room. But i found also smelt and minnow are much too small and i will rework all watch images at a later point in time.
Perfectly good enough for now. Good job with the clean transparency mask cut out.

beroset (approved):
Tested locally -- everything works.

